PR #53, opened 2014-10-26
put db access into background thread?

yulin2:
Hi, I'm doing research on performance for Android apps. I found some event handlers access db from UI thread, but Android docs suggest us to avoid such blocking calls in UI thread. Do they lead to any responsiveness issues?

I tried to refactoring by putting them into background tasks. Note that I only refactoring the places which won't introduce data races after using Thread/AsyncTask. Looking forward to see your comments.


grote:
Sorry, I never got around to apply this and now it is outdated and won't merge anymore.

PR #59, opened 2015-01-06
Add Dutch translation

TheLastProject:
The names of cities are largely untranslated, but aside from that, this should help Dutch people use the app more nicely. Keep up the great work, @grote! :)


grote:
Thanks a lot for this! Normally, I prefer Transifex for those, but for an initial translation, this will do.

Before I merge this, could you please remove the city strings you didn't translate?

grote:
Thanks a lot! Would be nice, if you'd advertise the next version in the Netherlands, to get more Dutch users.

Btw. Transifex also allows you to upload the .xml file.

PR #66, opened 2015-03-26
Fix public-transport-enabler URL

sim6:


grote:
Thanks a lot! :)

sim6:
Thank you!

PR #87, opened 2015-10-05
Minor layout changes

ByteHamster:


grote:
Awesome work! :)
Thanks for also improving Liberario. I really appreciate that!
How would you like to appear in the Contributor About section? As @ByteHamster or with real name?

grote:
Merged! One little issue I found with 42050d1 is that the popup menu in the favorite trip is dark in a lite theme. Any idea how to fix this?

ByteHamster:
> How would you like to appear in the Contributor About section?

That would be nice! As **ByteHamster** please.

> the popup menu in the favorite trip is dark in a lite theme.

I guess the theme is based on Light.**DarkActionBar** which also affects the menus. I have developed an app where I use a workaround for that - going to add a pull request soon.

PR #90, opened 2015-10-13
Changed favourites font color, added actionBarItemBackground, improved back navigation

ByteHamster:


grote:
Thanks a lot as usual! :)
This is now merged. I am looking forward to your next improvements which are always great!

PR #104, opened 2015-10-24
shared activity transition

ByteHamster:


grote:
Awesome, thanks a lot for fixing the transition! :)

grote:
I noticed a "jump" at the bottom of the card during the transition. I guess that is super difficult to fix, right?

ByteHamster:
> Why did you remove the CoordinatorLayout here? Did the transition not work with it?

I guess it's a bug in `CoordinatorLayout`:
After returning back to TripsActivity, the list scrolled randomly up or down by the height of the `Toolbar`.

> I noticed a "jump" at the bottom of the card during the transition. I guess that is super difficult to fix, right?

I tried to fix that but failed. Anyway, I think it looks much better with the animation ;)

grote:
Alright, thanks! Merged.

PR #108, opened 2015-10-26
Favourites

ByteHamster:
I'm getting a "Permission denied" error when trying to push, even if I *should* have push rights...
Does not matter, here is the commit ;)


grote:
Thanks a lot, this was a biggie! :) Merged!

PR #146, opened 2016-01-12
add VRN & VVS logos

dotlambda:
I just took the files from Wikimedia and resized them


grote:
Thanks a lot! Normally, I like transparent backgrounds, but here this will cause problems with dark/light themes. So it would be better to use a white background instead.

dotlambda:
sorry for that, I added white backgrounds

dotlambda:
By the way, how about mentioning that on https://transportr.grobox.de/#contribute?
I couldn't find a github repo of the site (how about setting one up?), otherwise I would have added it myself.

grote:
Thanks a lot! I am looking forward to your next contribution! :)

Btw. I added the background information to the homepage.

PR #151, opened 2016-01-24
Update Google Play badge URL

dotlambda:
I wasn't able to find a smaller version, so I had to resize it using HTML


grote:
Thanks for updating the badge! :)

Even better than resizing it with HTML would be to not load it from Google's server, but put it into the `/art` directory already re-sized. Btw. I prefer it to be smaller than the F-Droid logo ;)

dotlambda:
Now they have the same width, so the F-Droid badge is a little taller

grote:
Thanks a lot!

PR #157, opened 2016-02-15
Add Ontario (Canada) region.

chimo:
Hi!
As discussed, here's the PR for Ontario, Canada.

Please let me know if anything needs to be changed


grote:
Awesome, looks really good! :) Do you also want to add a logo, maybe the flag of Ontario or if there's one public transport company, use theirs?

chimo:
Looks like "the official 'symbols' of the Government of Ontario" can't be used without the [government's written authorization](https://www.ontario.ca/page/terms-use#section-5) (I'm guessing this includes the flag, but I could be wrong...), and there are more than one public transport company in that region.

Let me know if you have any ideas on how to proceed!

grote:
Wikipedia says that [the flag](https://en.wikipedia.org/wiki/Flag_of_Ontario#/media/File:Flag_of_Ontario.svg) and [the coat of arms](https://en.wikipedia.org/wiki/Coat_of_arms_of_Ontario#/media/File:Coat_of_Arms_of_Ontario.png) are in the public domain. But maybe it could just be a red Maple Leaf on white background with ON written in the bottom right corner? So other states of Canada could be added later using the same logo principle?

grote:
How about this?
network_ontario_logo

chimo:
Sorry for the delay; didn't get the chance to look at this yesterday.

Good idea, looks really good!
Should I include it in this PR or you prefer adding it yourself after? (I wouldn't want to take credit for your work)

grote:
Either way, I don't care ;)
Thanks a gain for your contribution! If you are interested in adding other providers, I'd be very happy about that! :)

chimo:
Logo added!

src/de/grobox/liberario/TransportNetworks.java
```diff
@@ -493,6 +493,16 @@ public TransportNetworks(Context context) {
 						.setStatus(TransportNetwork.Status.APLHA)
 		);
 
+		// Canada
+		region = getString(R.string.np_region_canada);
+
+		list.add(new TransportNetwork(context, NetworkId.ONTARIO)
+				         .setName("Ontario")
+				         .setDescription("Ottawa, Toronto")
```
grote: I just noticed after merging that these are not translatable strings. Are you sure these names are the same in every language?

Issue #144, opened 2016-01-12
Add support for Montreal (and other cities in Quebec)


baldurmen:
It looks to me that Montreal and quite a few other cities in the province of Quebec in Canada are on Navitia: https://navitia.opendatasoft.com/explore/dataset/ca-qc/table/

Having Montreal support would be great. Having all the cities supported by Navitia in Quebec would simply be awesome.

grote:
That would indeed be great. Would you like to help with that? Maybe have a look at what @kalon33 did in #137. The first step is to submit a PR to public-transport-enabler, then get it merged there. Afterwards, adding it to Transportr is fairly simple.

grote:
@chimo wouldn't this be a nice thing for you to do? ;)

chimo:
Sure! I'll probably have some free time for this tomorrow  :)

chimo:
Step 1: https://github.com/schildbach/public-transport-enabler/pull/76

baldurmen:
@chimo 's commit has been merged! What is the second step?

chimo:
2nd step would be to add it to transportr, kinda like this: https://github.com/grote/Transportr/commit/d8f4ccb2e392575d6596f45d4c87447658866107

anarcat:
anybody working on the next steps here?

chimo:
Yep, it's done. I'm guessing it'll be available in the next release. Or you can get the feature now if you compile it yourself.

anarcat:
oh i see, sorry, thanks.

grote:
I am working on some major refactorings at the moment. Once that is done, I will release a beta version in this repository: https://grobox.de/fdroid/
This will also include support for Quebec.

anarcat:
@grote please do let us know, can't wait to try th is out!

grote:
A pre-release has just been made available and can be obtained via F-Droid from his repo: https://grobox.de/fdroid/
Please file any issues you find with that release.

Issues specifically with Quebec data should be filed against public transport enabler directly.

anarcat:
@grote - i can confirm routing seems to work after checking a few connections. thanks, that is a great addition!

i did get a crash when fiddling around with the home location, but i do not know if it is related...?

grote:
Could you try to reproduce the crash and get a stack trace for me?

anarcat:
[...]

PR #165, opened 2016-03-12
Add Quebec (Canada) region.

chimo:


chimo:
(refs. #144)

baldurmen:
:D this is great.

I'm curious about naming the cities though. If you look at the data (https://navitia.opendatasoft.com/explore/dataset/ca-qc/table/), way more cities are supported than just Quebec, Montreal, Laval & Sherbrooke.

But then again, these are the only cities marked in "Geographic Area". Does it mean we'll still be able to see buses and trains from Chambly or in the Outaouais region even though cities won't be added?

grote:
Awesome! Looks good!

Have you seen the Ontario logo in the navigation drawer? The ON can't really be seen. I guess it will be similar with QC. Do you think it will still look good when the text is moved more towards the center?

@baldurmen you should be able to get results for all agencies listed on that page. If you want to know what exactly is in there, you can open the linked zip files and peek inside the data yourself or you wait for the next release to use Transportr to explore the data.

TLEM in England has a pretty long list of cities as well. I guess there's no reason to not add at least more of the bigger cities to the list if they are indeed supported. You could also describe it differently like "All major cities supported" (if that's true) or something similar. @chimo what do you think?

chimo:
> I'm curious about naming the cities though. If you look at the data (https://navitia.opendatasoft.com/explore/dataset/ca-qc/table/), way more cities are supported than just Quebec, Montreal, Laval & Sherbrooke.

@baldurmen  Ah yes, I didn't realize; I copied the (incomplete, turns out) list from here: https://navitia.opendatasoft.com/explore/dataset/ca-qc/ I'll add the others.

> Have you seen the Ontario logo in the navigation drawer? The ON can't really be seen. I guess it will be similar with QC. Do you think it will still look good when the text is moved more towards the center?

@grote Good catch, I'll try moving the text.

grote:
Thanks a lot! :)

Would you mind also changing the logo for Ontario and letting me know when you personally feel the status of those providers can be moved from `alpha` to `beta`?
I'm not living there and not using them, so for me it is hard to judge.

chimo:
Done in #167 !

I've been using the Ontario (well, Ottawa more specifically, I suppose) and I'm okay bumping it to beta.
Maybe @baldurmen can tell us about their experience with the Quebec region and if it's good we can do the same for that region?

baldurmen:
> I've been using the Ontario (well, Ottawa more specifically, I suppose) and I'm okay bumping it to beta.
> Maybe @baldurmen can tell us about their experience with the Quebec region and if it's good we can do the same for that region?

Hmmm, I can _try_ building it but I can't guaranty anything. I'll post
updates here.

grote:
@baldurmen you can also wait for the next release and let us know at any time.

baldurmen:
Hmm, I think I'll do that.

Issue #173, opened 2016-04-04
New network icons


nautilusx:
Hello,

here are some new or improved network icons.
![network_avv_logo](https://cloud.githubusercontent.com/assets/8766955/14259303/9c344b6a-faa6-11e5-9135-a02fca5f62fb.png)
![network_invg_logo](https://cloud.githubusercontent.com/assets/8766955/14259309/a1669106-faa6-11e5-93ee-3ba0aca1708f.png)
![network_mvg_logo](https://cloud.githubusercontent.com/assets/8766955/14259313/a48941e4-faa6-11e5-9038-6c6d3b47922a.png)
![network_sh_logo](https://cloud.githubusercontent.com/assets/8766955/14259317/a6bf7640-faa6-11e5-9c3d-c6f05806e232.png)
![network_vms_logo](https://cloud.githubusercontent.com/assets/8766955/14259318/a8a3319a-faa6-11e5-80f0-7007334f626f.png)
![network_vmv_logo](https://cloud.githubusercontent.com/assets/8766955/14259336/aa865c4e-faa6-11e5-9fa4-2ce777cc8d8b.png)
![network_vvo_logo](https://cloud.githubusercontent.com/assets/8766955/14259337/ac36a45e-faa6-11e5-9ca9-e88423f9358d.png)

There will be more in the next days ;o)

grote:
Nice, thanks a lot! :)

I'll wait for you to add the others here before adding them to the code.

What would actually be most awesome is to use vector graphics wherever possible. Android does not use standard SVG images, but there are tools to convert SVG into Android vector format:
https://inloop.github.io/svg2android/

We have the AVV logo here as SVG: https://github.com/grote/Transportr/blob/master/art/networks/avv.svg

nautilusx:
Hi Torsten,

the avv.svg has no white background. It looks not so nice if you use the dark theme.

I can make svg if you want but on your transportr homepage is the description for png.

grote:
Yes, I mention the white background on the homepage as well. You can easily add a white-background there with Inkscape.

The vector support on Android is new. When it works here, I will add this to the homepage. The challenge is to get a vector graphic in the format Android needs.

nautilusx:
Hi Torsten,

I made new svgs with inkscape and convert the svg with https://inloop.github.io/svg2android/

[svg_and_xml_logos.zip](https://github.com/grote/Transportr/files/203903/svg_and_xml_logos.zip)

But I get only new xml files to download with this converter.

[...]

PR #176, opened 2016-04-06
Added missing network logos

vinzv:
* Verkehrsgesellschaft Mecklenburg-Vorpommern
* Nahverkehrsverbund Schleswig-Holstein
* Verkehrsverbund Oberelbe
* Verkehrsverbund Mittelsachsen
* Märkische Verkehrsgesellschaft
* Nordhessischer Verkehrsverbund
* Nahverkehrsgesellschaft Baden-Württemberg
* Verkehrsverbund Vogtland
* Verkehrsmanagement-Gesellschaft Saar

Both PNG binaries and SVG sources included. All logos are public domain,
data taken from Wikipedia.


grote:
Awesome work @vinzv ! :)

We are just discussing using vector drawables instead of pngs in #173 where there are already some of the ones you did. Do you think you could turn your SVGs into vector drawables like described there and use them instead of the PNGs?

vinzv:
Just created them with SVG to VectorDrawable Converter and --lib BetterVectorDrawable. As I have no Android Dev Setup running here it's a blind shot. Hopefully the files work.

grote:
The "BetterVectorDrawable" lib won't work as this is a special library which I am not using. You don't need special parameters. There's only drawables with "hollow paths" like the middle of the letter "O" which can cause problems. There, the "--fix-fill-type" option might help.

grote:
Some more comments: All file names should be of the form `network_[NetworkId]_logo.[ext]`
Files with the extension `xml` need to be placed in `res/drawable` and files with `svg` should be in `art/networks`. When there is XML files for each logo, the PNG files are not needed anymore and should be deleted.

vinzv:
I re-created the XML files. The only warnings I got were:
`..
[Warning(s)] met.svg: Met unsupported element(s): rect.
..
[Warning(s)] network_avv_logo.svg: Met unsupported element(s): rect.
.
[Warning(s)] network_db_logo.svg: Met unsupported element(s): rect.
.......
[Warning(s)] network_vms_logo.svg: Met unsupported element(s): polygon.
....
[Warning(s)] ontario.svg: Met unsupported element(s): rect, text.`

grote:
The files that caused warning are having display problems. Usually, you should be able to solve that by transforming the problematic objects into paths. Inkscape -> Objects to Path.

As discussed in #173 the SVG files need an explicit white background which has exactly the same dimensions as the file. See the network_db_logo.svg as an example.

Sorry for all those comments, but we need to get this right, right? ;)

grote:
Also, looks like there's still PNGs in your PR which are also available as XML files. These need to be deleted, because the XML is not used, if there is a PNG.

grote:
One more thing ;)
The maximum size should be 64dp. This can be edited manually in the XML file. For quadratic images this means:
```xml
<vector xmlns:android="http://schemas.android.com/apk/res/android"
    android:width="64dp"
    android:height="64dp"
```

vinzv:
As I'm not on a Windows machine now and don't want to install Mono I used https://inloop.github.io/svg2android/ for conversion.
All XML files have 64dp now and a explicit white background. Hopefully the files work now.

grote:
Thanks a lot. This is looking great!

I just tried this out and noticed some small issues:
* You overwrote the existing and already correct `res/drawable/network_db_logo.xml` with something that has no background
* Many/All XML logos have duplicated attributes, such as `android:width` and `android:pathData`
* Many/All XML logos have no white background
* The invg logo xml is a broken DB logo
* MVG XML has no white background
* MVV XML is on the head and has no white background
* NAH SH still has a PNG, the XML is broken probably due to using the unsupported path fill option
* NVBW looks like it is the same as NAH SH, same for NVBW (I think the website might not always update the XML text)
* Ontario logo has no text, but the SVG is outdated, @chimo didn't update it when he changed the logo, I think.

I'll stop here, because I think many issues will be taken care of when converting the SVGs properly. Then, I'll check again.

grote:
I checked the SVGs and I think the reason why the white background isn't shown is because it used the unsupported fill rule.

There's an explanation here for how to fix this: https://github.com/a-student/SvgToVectorDrawableConverter#manual-editing
Basically, you need to select the icon in the top-right when filling areas:

Inkscape

PR #203, opened 2016-06-22
Store selected transport products in SharedPreferences [wip]

kas70:
These are my first steps. Thanks in advance for any help;)

### DIFF
--- a/src/de/grobox/liberario/Preferences.java
+++ b/src/de/grobox/liberario/Preferences.java
@@ -23,7 +23,13 @@ import android.preference.PreferenceManager;
 import android.support.annotation.Nullable;
 import android.util.Log;
 
+import java.io.Serializable;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.Set;
+
 import de.schildbach.pte.NetworkId;
+import de.schildbach.pte.dto.Product;
 
 public class Preferences {
 
@@ -35,6 +41,8 @@ public class Preferences {
 	public final static String WALK_SPEED = "pref_key_walk_speed";
 	public final static String OPTIMIZE = "pref_key_optimize";
 	public final static String EXIT_ON_BACK = "pref_key_exit_app_on_back_press";
+
+	private final static String SELECTED_PRODUCTS = "_selected_products";
 
 	public static String getNetwork(Context context, int i) {
 		SharedPreferences settings = context.getSharedPreferences(PREFS, Context.MODE_PRIVATE);
@@ -98,8 +106,47 @@ public class Preferences {
 		editor.commit();
 	}
 
+	public static EnumSet<Product> getProducts(Context context) {
+		// check if there's a preference for selected products for the current network
+		TransportNetwork current_network = getTransportNetwork(context);
+		String network_pref = current_network.getId().toString() + SELECTED_PRODUCTS;
+		SharedPreferences sp = context.getSharedPreferences(PREFS, Context.MODE_PRIVATE);
+		// get the selected products from the pref
+		Set<String> string_set = sp.getStringSet(network_pref, null);
+		// if there is no pref for the current network or all products are selected return all
+		if(string_set == null || string_set.size() == 0 || string_set.size() == Product.values().length){
+			return EnumSet.allOf(Product.class);
+		}
+		// if there is a pref, read the selected products and return them
+		else {
+			EnumSet<Product> products = EnumSet.noneOf(Product.class);
+			for(String s : string_set) {
+				products.add(Product.valueOf(s));
+			}
+			return products;
+		}
+	}
+
+	public static void setProducts(Context context, EnumSet<Product> products) {
+		if(products.size() == Product.values().length) return;
+		// get the current network
+		TransportNetwork current_network = getTransportNetwork(context);
+		String network_id = current_network.getId().toString();
+		SharedPreferences sp = context.getSharedPreferences(PREFS, Context.MODE_PRIVATE);
+		SharedPreferences.Editor editor = sp.edit();
+		String network_pref = network_id + SELECTED_PRODUCTS;
+		// store the selected products in a string set
+		Set<String> product_set = new HashSet<String>();
+		for(Product p : products) {
+			product_set.add(p.toString());
+		}
+		// set the pref
+		editor.putStringSet(network_pref, product_set);
+		editor.apply();
+	}
+
 	public static void setNetworkId(Context context, NetworkId id) {
-		SharedPreferences settings = context.getSharedPreferences(Preferences.PREFS, Context.MODE_PRIVATE);
+		SharedPreferences settings = context.getSharedPreferences(PREFS, Context.MODE_PRIVATE);
 		SharedPreferences.Editor editor = settings.edit();
 
 		String id1 = settings.getString("NetworkId", "");
--- a/src/de/grobox/liberario/fragments/DirectionsFragment.java
+++ b/src/de/grobox/liberario/fragments/DirectionsFragment.java
@@ -151,6 +151,7 @@ public class DirectionsFragment extends TransportrFragment implements TripHandle
 		productsAdapter = new FastItemAdapter<>();
 		ui.products.setHasFixedSize(true);
 		ui.products.setAdapter(productsAdapter);
+		products = Preferences.getProducts(getContext());
 		onProductsChanged(products);
 		ui.products.getBackground().setColorFilter(TransportrUtils.getButtonIconColor(getActivity()), SRC_ATOP);
 		ui.products.setOnTouchListener(new View.OnTouchListener() {
@@ -390,7 +391,7 @@ public class DirectionsFragment extends TransportrFragment implements TripHandle
 				pd.setCancelable(false);
 				pd.setIndeterminate(true);
 				pd.setButton(DialogInterface.BUTTON_NEGATIVE, "Cancel", new DialogInterface.OnClickListener() {
-					             @Override
+ 					             @Override
 					             public void onClick(DialogInterface dialog, int which) {
 						             mAfterGpsTask = null;
 						             dialog.dismiss();
--- a/src/de/grobox/liberario/fragments/ProductDialogFragment.java
+++ b/src/de/grobox/liberario/fragments/ProductDialogFragment.java
@@ -108,6 +108,7 @@ public class ProductDialogFragment extends DialogFragment {
 				// call listener if set
 				if(listener != null) {
 					EnumSet<Product> products = getProductsFromItems(adapter.getSelectedItems());
+					Preferences.setProducts(getContext(), products);
 					listener.onProductsChanged(products);
 				}
 				getDialog().cancel();


src/de/grobox/liberario/activities/MainActivity.java
```diff
@@ -205,6 +210,21 @@ public void onDrawerSlide(View drawerView, float slideOffset) {
...
+		super.onResume();
+		// check if there's a preference for selected products for the current network
+        TransportNetwork current_network = Preferences.getTransportNetwork(getContext());
+        SharedPreferences sp = getContext().getSharedPreferences("pref_selected_products", 0);
+        String network_pref = current_network + "_selected_products";
+        if(sp.contains(network_pref)){
+            // get the selected products from the pref
+            Set<String> product_set = sp.getStringSet(network_pref, new HashSet<String>());
+            //TODO: initiate the loading of the products from the product_set?
+            //where's actually the point, where the products are loaded?
```
grote: You don't need to touch `MainActivity` at all. Look into the `DirectionsFragment`. That's the only place where the products are used at the moment.

src/de/grobox/liberario/activities/MainActivity.java
```diff
@@ -205,6 +210,11 @@ public void onDrawerSlide(View drawerView, float slideOffset) {
 	}
 
 	@Override
+	protected void onResume() {
+		super.onResume();
+	}
+
+	@Override
```
grote: The changes in this file are not needed anymore ;)

src/de/grobox/liberario/fragments/ProductDialogFragment.java
```diff
@@ -108,7 +113,20 @@ public void onClick(View v) {
 				// call listener if set
 				if(listener != null) {
 					EnumSet<Product> products = getProductsFromItems(adapter.getSelectedItems());
-					listener.onProductsChanged(products);
+                    // get the current network
+                    TransportNetwork current_network = Preferences.getTransportNetwork(getContext());
+                    String network_pref = current_network + "_selected_products";
+                    SharedPreferences sp = getContext().getSharedPreferences("pref_selected_products", 0);
```
grote: Here, you are not using the constant `SELECTED_PRODUCTS` that you introduced.

Btw. I thought you create a shared-preferences file per network and in there have a `PRODUCTS` preference that you look for.

kas70: So i'll have to create  `	public final static String NETWORK_SELECTED_PRODUCTS = "pref_network_selected_products";`  for each network?

grote: No, if we do one shared preferences file per network that holds network specific settings, you could use something like `current_network.getId().toString()` for the name.

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -240,6 +244,23 @@ public void onResume() {
...
+            // TODO: convert Strings from product_set to Products
+            EnumSet<Product> h = EnumSet.noneOf(Product.class);
+            for(String s: product_set){
+                //h.add(??);
+            }
+			Serializable tmpProducts = h;
+            products = (EnumSet<Product>) tmpProducts;
+            onProductsChanged(products);
+		}
+
```
kas70: I have problems finding a way for putting the products from the StringSet (SharedPreference) into the EnumSet<Product>. Can you tell me how to do this?
If this is solved, i'll clean up my code and update the pull request.

grote: Something like this (untested):
```java
	public Set<String> productsToString(EnumSet<Product> products) {
		Set<String> stringSet = new HashSet<>(products.size());
		for(Product p : products) {
			stringSet.add(p.name());
		}
		return stringSet;
	}

	public EnumSet<Product> getProductsFromString(Set<String> stringSet) {
		EnumSet<Product> products = EnumSet.noneOf(Product.class);
		for(String s : stringSet) {
			products.add(Product.valueOf(s));
		}
		return products;
	}
```

kas70:
I think it's ready for a review.

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -240,6 +244,23 @@ public void onResume() {
 		displayFavouriteTrips();
 		refreshAutocomplete(false);
 
+		// check if there's a preference for selected products for the current network
+		TransportNetwork current_network = Preferences.getTransportNetwork(getContext());
+		String network_pref = current_network.getId().toString() + "_selected_products";
+		SharedPreferences sp = getContext().getSharedPreferences(Preferences.NETWORK_SETTINGS, 0);
```
grote: What is the `0` doing at the end?

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -240,6 +244,23 @@ public void onResume() {
...
+		// get the selected products from the pref
+		Set<String> string_set = sp.getStringSet(network_pref, new HashSet<String>());
+        // if there is a pref for the current network, read the selected products and update
+        if(string_set.size()>0) {
```
grote: I think it might be better to use `null` as the default parameter and check for `!= null` here. Then no `HashSet` needs to be initialized.

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -240,6 +244,23 @@ public void onResume() {
...
+			EnumSet<Product> products = EnumSet.noneOf(Product.class);
+			for (String s : string_set) {
+				products.add(Product.valueOf(s));
+			}
+			Serializable tmpProducts = products;
+			products = (EnumSet<Product>) tmpProducts;
+			onProductsChanged(products);
```
grote: Why you do this in `onResume()` again after it has been done in `onCreateView()`? Any reason to not change the code in `onCreateView()` directly and initialize the products there?

src/de/grobox/liberario/fragments/ProductDialogFragment.java
```diff
@@ -108,7 +110,28 @@ public void onClick(View v) {
 				// call listener if set
 				if(listener != null) {
 					EnumSet<Product> products = getProductsFromItems(adapter.getSelectedItems());
-					listener.onProductsChanged(products);
+                    // get the current network
+                    TransportNetwork current_network = Preferences.getTransportNetwork(getContext());
+					String network_id = current_network.getId().toString();
+                    SharedPreferences sp = getContext().getSharedPreferences(Preferences.NETWORK_SETTINGS, 0);
```
grote: Same question for the `0` here.

src/de/grobox/liberario/fragments/ProductDialogFragment.java
```diff
@@ -108,7 +110,28 @@ public void onClick(View v) {
...
+                    SharedPreferences.Editor editor = sp.edit();
+                    String network_pref = network_id + "_selected_products";
+					if(products.size()==Product.values().length){
+                        // remove pref if all products are selected
+                        editor.remove(network_pref);
+                        editor.apply();
```
grote: I think we don't need to remove the preferences with the new approach. It is sufficient to only store them when the user changes them and otherwise leave them like they are.

grote:
This is looking good already. Maybe to clean up the code further, it would be nice to put all the retrieval and store to the `Preferences` class into a static method. Something like:
```java
EnumSet<Product> getProducts();
setProducts(EnumSet<Product> products);
```
Then, you just need to call those methods in the fragments.

src/de/grobox/liberario/fragments/DirectionsFragment.java
kas70: From the API-Docs:
`mode 	int: Operating mode. Use 0 or MODE_PRIVATE for the default operation.  
File creation mode: the default mode, where the created file can only be accessed by the calling application (or all applications sharing the same user ID). `

src/de/grobox/liberario/fragments/DirectionsFragment.java
kas70: Done.

src/de/grobox/liberario/fragments/ProductDialogFragment.java
kas70: See above.

src/de/grobox/liberario/fragments/ProductDialogFragment.java
kas70: Ok, i'll remove those lines.

kas70:
Ouch! Due to some reformatting the changes look ugly.

src/de/grobox/liberario/fragments/DirectionsFragment.java
grote: Ok thanks in this case, I prefer `MODE_PRIVATE` because the meaning is clearer when just looking at it than just `0`.

grote:
> Ouch! Due to some reformatting the changes look ugly.

Indeed :( Can you undo that? Do you use Android Studio? There should be a code-style file in the repository and used automatically.

kas70:
I'll try.

kas70:
@grote Should be fine now.

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -390,7 +391,7 @@ private void search() {
 				pd.setCancelable(false);
 				pd.setIndeterminate(true);
 				pd.setButton(DialogInterface.BUTTON_NEGATIVE, "Cancel", new DialogInterface.OnClickListener() {
-					             @Override
+ 					             @Override
```
grote: Thanks, the diff looks much better now! :)
This here is the only one you missed.

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
 		editor.commit();
 	}
 
+	public static EnumSet<Product> getProducts(Context cont) {
+		// check if there's a preference for selected products for the current network
+		TransportNetwork current_network = Preferences.getTransportNetwork(cont);
```
grote: I think in this class you can just write `getTransportNetwork(cont)`.

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+		String network_pref = current_network.getId().toString() + "_selected_products";
+		SharedPreferences sp = cont.getSharedPreferences(Preferences.NETWORK_SETTINGS, Context.MODE_PRIVATE);
```
grote: So you are just using one preferences file to store preferences for *all* networks. That is fine as well, but in this case, you can just re-use the existing `PREFS` constant and you don't need your own. Also no need anymore to prefix with `Preferences.`.

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+		// get the selected products from the pref
+		Set<String> string_set = sp.getStringSet(network_pref, null);
+		// if there is a pref for the current network, read the selected products and update
+		if(string_set != null) {
+			EnumSet<Product> products = EnumSet.noneOf(Product.class);
+			for(String s : string_set) {
+				products.add(Product.valueOf(s));
+			}
+			Serializable tmpProducts = products;
+			products = (EnumSet<Product>) tmpProducts;
```
grote: Why do you need `tmpProducts` here? Wouldn't it work without it?

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+			return products;
+		} else {
+			return EnumSet.noneOf(Product.class);
+		}
+	}
+
+	public static void setProducts(Context cont, EnumSet<Product> products) {
+		// get the current network
+		TransportNetwork current_network = Preferences.getTransportNetwork(cont);
```
grote: The `Preferences.` prefix can be removed here as well.

grote: Oh and maybe just to stay consistent with how it is done everywhere else, use `context` instead of `cont`.

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+		if(string_set != null) {
+			EnumSet<Product> products = EnumSet.noneOf(Product.class);
+			for(String s : string_set) {
+				products.add(Product.valueOf(s));
+			}
+			Serializable tmpProducts = products;
+			products = (EnumSet<Product>) tmpProducts;
+			return products;
+		} else {
+			return EnumSet.noneOf(Product.class);
```
grote: Why don't you return no products? Wouldn't it be better to return *all* products in this case?

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+		}
+	}
+
+	public static void setProducts(Context cont, EnumSet<Product> products) {
+		// get the current network
```
grote: Before this even happens, you could check if `products` is equal to *all* products and if so just return here without saving anything. (This would require returning all products above)

src/de/grobox/liberario/Preferences.java
```diff
@@ -98,6 +105,44 @@ public static void setPref(Context context, String pref, boolean value) {
...
+		TransportNetwork current_network = Preferences.getTransportNetwork(cont);
+		String network_id = current_network.getId().toString();
+		SharedPreferences sp = cont.getSharedPreferences(Preferences.NETWORK_SETTINGS, Context.MODE_PRIVATE);
+		SharedPreferences.Editor editor = sp.edit();
+		String network_pref = network_id + "_selected_products";
```
grote: Could you please turn `"_selected_products"` here and above in the getter into a constant?

grote:
This is looking really good now! :) Yes I still had a lot of comments, but this is all small stuff and should be the last ones now. Looking forward to merge this. Planning a new release this WE.

grote:
Oh it would be awesome if you could squash all your commits into one. Thanks a lot! :)

src/de/grobox/liberario/fragments/DirectionsFragment.java
kas70: Can't make out the difference here. Looks like a space. But i can't figure it out in Android Studio.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Done.

src/de/grobox/liberario/Preferences.java
kas70: Does it make sense to load all products the next time the Fragment loads if a user deselected all before?

kas70:
If there are no objections from your site, i'll squash them all into one commit.

src/de/grobox/liberario/Preferences.java
grote: Deselecting all products doesn't make sense (should maybe not even possible), but if the user deselected some, next time the fragment loads, the pref has been set and should be used, so `string_set` should not be `null` anymore.

grote:
Thanks a lot for making another round of changes! :)
No harm in squashing already. It looks all good now, but I'll test the code practically before merging.

kas70:
Done. Thank you for your advice! It was fun working on this 😉 Looking forward to the release!

grote:
Merged! Works nicely, good work! :)

I am looking forward to your next contribution :)

grote:
Meh there was a nasty crash bug in your code which I only found *after* I did the release already :(
https://github.com/grote/Transportr/commit/2dd9b885363ec6c06851513c2ed460c82d6d7d70

kas70:
Sorry to hear that :/ So the problem was, that i haven't checked `getTransportNetwork` for returning null?!

grote:
Yes that was the problem, because it is `null` for the very first start before the user selected one.

PR #210, opened 2016-07-04
Prevent users from deselecting all transport products

kas70:


grote:
Thanks, looks good! Just let me test this practically before merging.

PR #226, opened 2016-08-11
Fix #225

### LINKED ISSUE #225: Selected transportation modes are not saved correctly
Heyho,

I'm running v1.1.3 from FDroid. I disabled subways in the transportation mode selection once and since then it is always disabled on app-startup, even if I enabled it again in the previous session. Also in the transportation mode selection popup the checkboxes are not labeled, there is just empty space, no text left of them. Let me know if you need a screenshot.

schachmat:
DISCLAIMER: I did not test the change, but it should work.


grote:
Thanks a lot, merged! :)

PR #239, opened 2016-09-30
Improve network provider list

### LINKED ISSUE #234: Some keys uses internal ID instead of a string
From https://www.transifex.com/grobox/transportr/translate/#it/strings/39245513?reviewed=no

Question by @airon90
> Why it [the key "WIENER LINIEN"] is written in caps lock? Shouldn't it be "Wiener Linien" instead of "WIENER LINIEN"?

Answer by @grote
> There's also some other's where the name matches the internal ID rather than the geographical regions where the provider is active. Feel free to open an issue on GitHub about this as this has to be addressed in the code.

patkan:
* I added emoji-flags to every region.
 * I moved the flags before the names of the regions. Downturn of this is that the sorting is now different. Benefit is that it looks cleaner imho.
 * I introduced helper-functions for the structuring of the strings
 * I sorted the string-xml
 * I reviewed every provider and tried to update/improve it and correct data where necessary.
 * I deactivated Israel since it seems to be broken.
 * I put all strings into the xml, because I felt it is cleaner this way.

This should also be relevant to #234, but it does not introduce a string for every network, (I used strings only where the name differed from the network name. If you want to I can also add strings for those cases.)

From my standpoint this can be merged, however I am also open for discussion. If anything should be changed I can rebase my PR.

Apart from that, I fear this will mess with transifex. I made some strings that where translatable (and still have translations) non-translatable and swapped one or two strings (name with desc). I did not touch the translated string-files since I do not know the translation-workflow.

### DIFF
--- a/res/drawable-xhdpi/.directory
+++ b/res/drawable-xhdpi/.directory
@@ -1,6 +1,0 @@
-[Dolphin]
-HeaderColumnWidths=270,74,108
-SortRole=date
-Timestamp=2016,9,30,23,17,15
-Version=3
-ViewMode=1
--- a/res/drawable-xhdpi/.directory
+++ b/res/drawable-xhdpi/.directory
@@ -1,0 +1,6 @@
+[Dolphin]
+HeaderColumnWidths=270,74,108
+SortRole=date
+Timestamp=2016,9,30,23,17,15
+Version=3
+ViewMode=1
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -204,7 +204,7 @@ there are any).</string>
     <string name="np_desc_wien">Vienna</string>
 
     <string name="np_region_belgium">Belgium</string>
-    <string name="np_desc_sncb">trains in Belgium</string>
+    <string name="np_desc_sncb">Trains in Belgium</string>
 
     <string name="np_region_br">Brazil</string>
     <string name="np_name_br">Brazil</string>
@@ -226,7 +226,7 @@ there are any).</string>
 
     <string name="np_region_europe">Europe</string>
     <string name="np_name_rt">Europe</string>
-    <string name="np_desc_rt">long-distance trains only</string>
+    <string name="np_desc_rt">Long-distance trains only</string>
     <string name="np_desc_rt_networks">Railteam alliance</string>
 
     <string name="np_region_finland">Finland</string>
@@ -289,7 +289,6 @@ there are any).</string>
     <string name="np_desc_vrn">Baden-Württemberg, Rheinland-Pfalz, Mannheim, Mainz, Trier</string>
     <string name="np_desc_vrr">North Rhine-Westphalia, Köln, Bonn, Essen, Dortmund, Düsseldorf, Münster, Paderborn, Höxter</string>
     <string name="np_desc_vrs">Cologne, Bonn</string>
-    <string name="np_name_vvm" translatable="false">move</string>
     <string name="np_desc_vvm">Swabia, Mittelschwaben, Krumbach, Günzburg, Memmingen</string>
     <string name="np_desc_vvo">Saxony, Dresden</string>
     <string name="np_desc_vvs">Baden-Württemberg, Stuttgart</string>
@@ -337,13 +336,11 @@ there are any).</string>
 
     <string name="np_region_sweden">Sweden</string>
     <string name="np_desc_se">Sweden, Stockholm</string>
-    <string name="np_name_stockholm">Stockholm</string>
-    <string name="np_desc_stockholm">Stockholm</string>
 
     <string name="np_region_switzerland">Switzerland</string>
     <!-- SBB is german/international, CFF is french and FFS is italian! So this IS translatable. -->
     <string name="np_name_sbb" translatable="true">SBB</string>
-    <string name="np_desc_sbb">trains in Switzerland</string>
+    <string name="np_desc_sbb">Trains in Switzerland</string>
     <string name="np_desc_vbl">Luzern</string>
     <string name="np_desc_zvv">Zurich</string>
 
@@ -354,7 +351,7 @@ there are any).</string>
     <string name="np_region_usa">United States of America</string>
     <string name="np_name_rtachicago">RTA</string>
     <string name="np_desc_rtachicago">Chicago</string>
-    <string name="np_desc_rtachicago_networks" translatable="false">cta, Metra, pace</string>
+    <string name="np_desc_rtachicago_networks" translatable="false">CTA, Metra, Pace</string>
     <string name="np_desc_septa">Bucks, Chester, Delaware, Montgomery and Philadelphia Counties</string>
     <string name="np_name_sf">San Francisco Bay Area</string>
     <string name="np_desc_sf">Alameda, Contra Costa, Marin, Napa, Solano, Sonoma, San Francisco, San Mateo and Santa Clara Counties</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -183,134 +183,152 @@ there are any).</string>
     <string name="changelog_title">What\'s New</string>
     <string name="changelog_show_full">More…</string>
 
+    <string name="np_region_australia">Australia</string>
+    <string name="np_desc_met">Australia, Victoria, Melbourne</string>
+    <string name="np_name_sydney">Sydney</string>
+    <string name="np_desc_sydney">Australia, New South Wales, Sydney</string>
+
+    <string name="np_region_austria">Austria</string>
+    <string name="np_desc_ivb">Innsbruck</string>
+    <string name="np_desc_linz">Upper Austria, Linz</string>
+    <string name="np_desc_oebb">Austria</string>
+    <string name="np_desc_stv">Steiermark, Graz, Marburg, Maribor</string>
+    <string name="np_desc_svv">Salzburg</string>
+    <string name="np_desc_vor">Lower Austria, Burgenland, Wien</string>
+    <string name="np_desc_vvt">Tirol</string>
+    <string name="np_name_wien">Vienna</string>
+    <string name="np_desc_wien">WIENER LINIEN</string>
+
+    <string name="np_region_belgium">Belgium</string>
+
+    <string name="np_region_br">Brazil</string>
+    <string name="np_name_br">Brazil</string>
+    <string name="np_desc_br">São Paulo, Rio de Janeiro, Belo Horizonte and Porto Alegre</string>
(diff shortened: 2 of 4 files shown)


src/de/grobox/liberario/TransportNetworks.java
```diff
@@ -579,4 +610,12 @@ TransportNetwork getTransportNetwork(String idString) {
 	private String getString(int res) {
 		return context.getString(res);
 	}
+
+	private String region(String name, String flag) {
+		return flag + " " + name;
+	}
+
+	private String description(String desc, String networks) {
```
grote: Nice idea to have helper methods for that!

You could even improve that further like so: Except for `flag` all those could take a `@StringRes int` type as parameter, so you can remove all the `getString()` methods above and use them here instead.

patkan: That's a good point, thank you!
I did this in an earlier version, however I changed that. That was after I saw the feature of Android-Studio that it can "autocomplete" the strings. (It shows previews like in the screenshot)
screen
The benefit of using directly the ints is of course less boilerplate-code.

I am not sure and I see benefits in both. Just tell me what you like more and I will patch it if needed.

patkan: I thought a bit more about this:

- This is only helpful for users of Android-Studio (or editors that do that visual replacement)
- When in use, it is a nice feature. But normally the strings are always formed the same and the benefit is not very large.
- For those where the editor does not do this, this is just more boilerplate.

So I think I would prefer now the option that you proposed. I have added a commit to this PR.

grote: Did you try using the `@StringRes` annotation for the parameter like I had said? The text expansion should then work as before. Also, looks like you forget to call `getString()` now. Did you test your latest commits?

patkan: Sorry, I overlooked the annotation when reading your response. I did test it, however I messed up my commit :-1:  The last commit should have fixed that now.

res/values/strings.xml
```diff
@@ -183,133 +183,183 @@ there are any).</string>
...
     <string name="np_desc_vms">Mittelsachsen, Chemnitz</string>
-    <string name="np_desc_nasa">Saxony, Leipzig, Sachsen-Anhalt, Magdeburg, Halle</string>
-    <string name="np_desc_vrr">North Rhine-Westphalia, Köln, Bonn, Essen, Dortmund, Düsseldorf, Münster, Paderborn, Höxter</string>
-    <string name="np_desc_mvg">Märkischer Kreis, Lüdenscheid</string>
-    <string name="np_desc_nvv">Hesse, Frankfurt am Main, Kassel</string>
+    <string name="np_desc_vmv">Mecklenburg-Vorpommern, Schwerin</string>
     <string name="np_desc_vrn">Baden-Württemberg, Rheinland-Pfalz, Mannheim, Mainz, Trier</string>
+    <string name="np_desc_vrr">North Rhine-Westphalia, Köln, Bonn, Essen, Dortmund, Düsseldorf, Münster, Paderborn, Höxter</string>
+    <string name="np_desc_vrs">Cologne, Bonn</string>
+    <string name="np_name_vvm" translatable="false">move</string>
```
grote: What is `move`?

res/values/strings.xml
```diff
@@ -183,133 +183,183 @@ there are any).</string>
...
+    <string name="np_name_spain">Spain</string>
+    <string name="np_desc_spain">Madrid buses, Barcelona, Basque</string>
+    <string name="np_desc_spain_networks" translatable="false">EMT, TMB, Moveuskadi</string>
 
-    <string name="np_desc_ns">Netherlands, Amsterdam</string>
-    <string name="np_desc_dsb">Denmark, Copenhagen</string>
+    <string name="np_region_sweden">Sweden</string>
     <string name="np_desc_se">Sweden, Stockholm</string>
+    <string name="np_name_stockholm">Stockholm</string>
     <string name="np_desc_stockholm">Stockholm</string>
```
grote: I think we can delete the Stockholm strings and the provider. Doesn't look like it is coming back.

res/values/strings.xml
```diff
@@ -183,133 +183,183 @@ there are any).</string>
...
-    <string name="np_desc_tlem">Greater London, Birmingham, Derbyshire, Leicestershire, Rutland, Northamptonshire, Nottinghamshire, Lincolnshire, Berkshire, Buckinghamshire, East Sussex, Hampshire, Isle of Wight, Kent, Oxfordshire, Surrey, West Sussex, Essex, Hertfordshire, Bedfordshire, Cambridgeshire, Norfolk, Suffolk, Somerset, Gloucestershire, Wiltshire, Dorset, Devon, Cornwall, West Devon, Stowford, Eastleigh, Swindon, Gloucester, Plymouth, Torbay, Bournemouth, Poole</string>
-    <string name="np_desc_tlwm" tools:keep="@string/np_desc_tlwm">Birmingham</string>
-    <string name="np_desc_tfi">Ireland, Dublin, Belfast</string>
-    <string name="np_desc_atc">Emilia-Romagna, Bologna</string>
-    <string name="np_desc_pl">Poland, railways in Poland</string>
+
+    <string name="np_region_switzerland">Switzerland</string>
+    <!-- SBB is german/international, CFF is french and FFS is italian! So this IS translatable. -->
+    <string name="np_name_sbb" translatable="true">SBB</string>
+    <string name="np_desc_sbb">trains in Switzerland</string>
```
grote: `Trains`

res/values/strings.xml
```diff
@@ -183,133 +183,183 @@ there are any).</string>
...
-    <string name="np_name_bayern">BAVARIA</string>
-    <string name="np_name_paris">PARIS</string>
-    <string name="np_name_sydney">SYDNEY</string>
-    <string name="np_name_stockholm">STOCKHOLM</string>
-    <string name="np_name_wien">VIENNA</string>
-    <string name="np_name_spain">Spain</string>
+    <string name="np_region_usa">United States of America</string>
+    <string name="np_name_rtachicago">RTA</string>
+    <string name="np_desc_rtachicago">Chicago</string>
+    <string name="np_desc_rtachicago_networks" translatable="false">cta, Metra, pace</string>
```
grote: Are you sure that `cta` and `pace` are both lowercase? If this comes from Navitia, don't trust them to have this right ;)

src/de/grobox/liberario/TransportNetworks.java
grote: Thanks for fixing. Now it does auto-expand the strings, right? ;)

I had yet another idea (sorry for that :stuck_out_tongue:): you don't actually need the description helper method. This could be another `setDescription()` method of the `TransportNetwork` that takes two arguments. Would be more elegant, but not necessary to do it like this.

src/de/grobox/liberario/TransportNetworks.java
```diff
@@ -579,4 +611,12 @@ TransportNetwork getTransportNetwork(String idString) {
 	private String getString(int res) {
 		return context.getString(res);
 	}
+
+	private String region(@StringRes int name, String flag) {
+		return flag + " " + getString(name);
```
grote: I'm not sure I like the messed up order this causes although I agree that flags look better in the front.

res/values/strings.xml
patkan: It's MObilitätsVErbund. I followed the API-link in the provider to http://mobilitaetsverbund.de. They have "move" as something like a logo.
I don't know the provider myself so I am not sure how this has to be correctly.

grote: So this is not http://vvm-info.de/home/startseite/startseite.jsp since the provider is called VVM internally?

grote: It might be that move is just providing a service to VVM by providing the API.

res/values/strings.xml
patkan: I took that from the logos on the website http://rtachicago.com/.
However on closer examination they tend to write Pace and CTA. I think I will change it to this style.

src/de/grobox/liberario/TransportNetworks.java
patkan: You are right, that's way more elegant. I will change this.

res/values/strings.xml
patkan: Ah ok, I think I misunderstood this then. I will have a closer look at it.

src/de/grobox/liberario/TransportNetworks.java
patkan: Could we maybe sort by another string? Then we could sort by the name without flags and display with flags.

grote: Yes, there's several possibilities for how to do this. The flag doesn't need to be part of the same string. It could be two strings that are combined on view creation in the adapter. But that's out of scope of this MR. Feel free to work on this afterwards.

res/drawable-xhdpi/.directory
```diff
@@ -0,0 +1,6 @@
+[Dolphin]
```
patkan: argh, I will remove this :-/

patkan:
OK, I think I have now everything adressed. (If there is still anything, I am happy to fix it.)
Before merging I would squash the last two commits together (just for cosmetical reasons). Just tell me when this is ready.

src/de/grobox/liberario/TransportNetwork.java
```diff
@@ -57,6 +57,12 @@ public TransportNetwork setDescription(@Nullable String description) {
 		return this;
 	}
 
+	public TransportNetwork setDescription(@Nullable String description, @Nullable String networks) {
```
grote: Do we ever need to pass in `null`? If yes, we should keep the annotation. 

Is it ever a problem if we pass in `null`, but we never need to do it? If yes, it should have a `@NonNull` annotation.

patkan: What happens when the current language has no translation? Does the context then contain simply the original string?
I think both variables do not need to be Nullable. When we pass on null it just looks awkward, but nothing "bad" happens.

grote: It will use the English string. So you could remove the annotation.

Btw. you can also squash commits yourself and force push into your branch. But when merging they will all be squashed into one anyway now.

patkan:
OK, I meant that you should tell me when you are ready, so that I can squash without interfering with your activities.
But if you squash on merging than I do not need to do it. So this PR is good to go from my point.

grote:
Merged! Thanks a lot! This was great work! :)

patkan:
You're welcome. I had a good and helpful guide ;-)

Issue #238, opened 2016-09-30
Make network logos the same size


patkan:
The transport network logos in the picker have different sizes. I think they should all be the same size.
How can this be accomplished? Simply rescaling the svg and then putting it through the exporter? I tried it, but it seemed to not always work.
And if they should all be the same size, then how large should they be?

grote:
SVG logos should all have 64dpx64dp. If they don't have that, this can easily be changed in their XML file.

I suspect this issue to be really about updating the remaining raster graphics to SVG. Could that be it? If so,  @nautilusx and @vinzv had started with that, but I haven't seen more PRs from them so far.

patkan:
And now I understand why some networks have magically a logo even when I could not find it. Yep, the problem is that the "false" sized networks PNGs are.

grote:
I updated the title of this issue to reflect that it is really about changing the graphics to vector drawables. The homepage has information on how to do that here: https://transportr.grobox.de/#contribute

nautilusx:
It's on my to-do-list for October. I will create/update all other logos. Germany and Austria are done.

patkan:
On my way through the transport networks I found the new logo for TFI in Ireland and created it. It's not perfect but if you start working on the logos I could give you this one. Alternatively feel free to pull it in, I have it in my fork of this repo.

grote:
@patkan Please feel free to open a PR just for one logo if you have it already. What's done is done.

grote:
@nautilusx Any news?

nautilusx:
Here are the next 6 Logos.
[6xml.zip](https://github.com/grote/Transportr/files/706778/6xml.zip)

grote:
Thanks a lot @nautilusx! Added in a1d509c0319ba45b8ef14732ab87dc83d81cea01.

grote:
New link is now here: https://transportr.app/contribute/#design

PR #241, opened 2016-10-01
update TFI logo, rename FR-SW logo template

patkan:
In relation to #238.
I saw that the logo of Transport for Ireland has changed. (They have a fancy new website and such: http://www.transportforireland.ie/)
So I tried to recreate it as svg. I did not find the exact font, so this is a bit off. I think it is enough for recognizing, but if someone gets it better it's also ok ;-)


grote:
This image here should be large enough to trace it in inkscape: https://www.luas.ie/assets/files/RPA%20Marketing%20Campaign/TFI%20Tile%20and%20Logotype.png so you don't need to use a different font.

patkan:
Thanks for the link, I updated my commit. The logo should now look the same as the original.

grote:
Thanks for this PR and the quick update! :)

PR #242, opened 2016-10-02
fix spelling of Hanover

patkan:
as requested in transifex


patkan:
You're welcome. I would like to fix the network picker list, but I don't think I will have the time to do it soon. So don't wait for me.

Issue #71, opened 2015-04-27
Quick to home


DJaeger:
There should be an possibility to quickly start routing from current possition at current time to home.

grote:
Thanks for the proposal. I wouldn't know where to place this in the UI, but it could be a home screen widget that directly does the query for trips.

Pull are requests very welcome!

DJaeger:
Or an home (for example an house) button on start sceen (maybe top bar?) directly under the search button.

grote:
I want to keep the start screen as simple and concise as possible and I'm afraid it is not immediately clear what this button does. So I would prefer the home screen widget option. Here's how this could be done: https://developer.android.com/guide/topics/appwidgets/index.html

ertaiNL:
I'll look into this

grote:
Awesome! Let me know if you need some help, for example with a widget icon.

grote:
@ertaiNL Any news?

patkan:
@grote Do you have any proposals for a widget layout? I think it should

* contain some visual indicator that it belongs to Transportr (e.g. the logo)
* the configured target (should be configured once for the widget upon creation on the homescreen)
* the whole widget should be clickable and trigger the Activity

patkan:
And how should the widget handle the Transport-Networks?
Should it upon configuration ask for a network and then set it every time the widget is used? Or should it just try to navigate in the currently active network? I think for the long run the first solution would be nice, but the latter is easier to implement.

patkan:
And one last thing: should the search for the route directly start or should the search-dialog (MainActivity) be just pre-populated with the relevant from and to-data?

grote:
Are you interested in implementing this feature @patkan? If so, here's how I imagine this to be done:
* The widget has the size of a normal app icon and uses a logo in the style of the transportr logo: white house on red background. A placeholder image can be used initially.
* The initial implementation should be as simple as possible, so no configuration at all.
* When the widget is clicked, an intent is send to the `MainActivity` that fills out the form with the current location and the target with the home location.
* If there is no existing home location, the home chooser fragment should be opened
* As soon as the current position has been found and a valid home location has been entered, the search should start automatically.
* The widget should work within the currently selected transport network
* There should be no need to directly use the AmbigousLocationActivity

[...]

PR #253, opened 2016-11-01
Add quickhome widget

patkan:
Since this is easier here for you to comment, I create a PR in succession to #71. I have got it more or less working. The graphical design of the widget is still to be done but apart from that all functionality should be basically there. I still have problems to properly populate the form after setting the HOME-location, but the standard usecase (searching with a set HOME-location) seems to work.

I am happy about every form of feedback, feel free to tear my PR apart ;-)


res/values-v14/dimens.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+
+    <dimen name="widget_margin">0dp</dimen>
```
grote: Why does this need a different padding on API14+? Do you have a link to the documentation for this?
Also, this is used as padding, not margin, so the name should reflect that.

res/values/strings.xml
```diff
@@ -185,6 +185,8 @@ there are any).</string>
     <string name="changelog_title">What\'s New</string>
     <string name="changelog_show_full">More…</string>
 
+    <string name="widget_name_quickhome">Transportr: get home</string>
```
grote: style nitpick: maybe `Get Home` or even `Get Me Home`?

res/xml/quickhome_widget_info.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<appwidget-provider xmlns:android="http://schemas.android.com/apk/res/android"
+                    android:minWidth="40dp"
+                    android:minHeight="40dp"
+                    android:updatePeriodMillis="86400000"
```
grote: I am not experienced with widgets. What is this `updatePeriodMillis` for?

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -426,9 +435,42 @@ private void processIntent() {
...
+						Location home = RecentsDB.getHome(getContext());
+						if(home != null) {
+							to = home;
+						} else {
+							// show home picker dialog
+							changingHome = true;
+							HomePickerDialogFragment setHomeFragment = HomePickerDialogFragment.newInstance();
+							setHomeFragment.setOnHomeChangedListener(this);
+							FragmentTransaction ft = getActivity().getSupportFragmentManager().beginTransaction();
+							setHomeFragment.show(ft, HomePickerDialogFragment.TAG);
```
grote: The indentation is getting crazy here. Maybe this whole part get be factored out into two methods. one that processes the intent and the other showing the HomePickerDialogFragment if necessary. But I am not sure, we need this HomePickerDialogFragment stuff here.

A first idea is to maybe have a `LocationView#setLocation()` method that takes directly a `WrapLocation` and then handles everything based on those wrap locations on its own like it is now.

res/values-v14/dimens.xml
patkan: I have found an answer on [stackoverflow](http://stackoverflow.com/a/22295535/4244236) and in the [Android Documentation](https://developer.android.com/guide/topics/appwidgets/index.html#CreatingLayout). The reason is that API14+ adds a margin on its own and behaves thus differently. The name margin was used in the documentation, so I copied it.

res/values/strings.xml
patkan: You are right. I will change that.

res/xml/quickhome_widget_info.xml
grote: Better set it to `0` I'd say.

res/values-v14/dimens.xml
patkan: But apart from that the whole alignment of the widget is still off. :-) I will have to see how to do that properly.

src/de/grobox/liberario/fragments/DirectionsFragment.java
patkan: Thank you, I will look into that. Definitely, you are right. The indentation is a bit much.

patkan:
Thank you @grote for the feedback. I would be happy if you could design a graphic for the widget, since I am not that good at that. I think you even should be able to push into this branch. (Or just attach it here or something)

grote:
I now got around to try it. Looks nice! It even works when no home location has been set before. But there's an issue when GPS is not enabled. Have you tried that?

I'll design an icon and send it to you, then we can look into the widget positioning issues. Ideally, the widget doesn't take up more space of the home screen than app icons.

patkan:
> It even works when no home location has been set before.
There sometimes seem to be issues like the search is not starting. I haven't yet managed to observe it reproducibaly. Maybe there is something like a race-condition.
> But there's an issue when GPS is not enabled. Have you tried that?
Oh yeah, I haven't thought of that. After having that GPS-problem the app seems to have problems with the intents getting out or something like this.

The best way is probably to move the relevant parts of the code into the LocationView and then check again for errors. I've got the impression that this might solve at least some quirks.

grote:
How about this as an image for the widget?

home

You should be able to [download it here](https://romannurik.github.io/AndroidAssetStudio/icons-launcher.html#foreground.type=clipart&foreground.space.trim=1&foreground.space.pad=0.15&foreground.clipart=res%2Fclipart%2Ficons%2Faction_home.svg&foreColor=fff%2C0&crop=0&backgroundShape=square&backColor=f24236%2C100&effects=shadow).

patkan:
I like it :-) Cool app... I managed to download the files
If we later need a SVG, we would have to redraw it, right?

grote:
SVG for Android XML is not possible due to the shadow. The house itself is available as SVG, the rest not.

grote:
@patkan do you consider this ready? If not, what is missing?

patkan:
Unfortunately not yet :-(
The widget is still not centered when I place it.
My refactoring is not completely done since I would like to structure it a bit better.

I will try to take some time in the coming days to finish it or at least properly write down what is still to be done.

res/layout/widget_quickhome.xml
```diff
@@ -0,0 +1,17 @@
...
+             android:layout_height="match_parent"
+             android:id="@+id/widget_quickhome_frame">
+
+    <ImageView
+        android:layout_width="64dp"
+        android:layout_height="64dp"
+        android:layout_gravity="center_horizontal"
+        app:srcCompat="@drawable/ic_quickhome_widget"
+        android:id="@+id/widget_quickhome_logo"
+        android:background="@drawable/ic_quickhome_widget"/>
```
grote: Are you intentionally using `background` instead of `src` here?

src/de/grobox/liberario/QuickhomeWidget.java
```diff
@@ -0,0 +1,65 @@
+/*    Transportr
+ *    Copyright (C) 2013 - 2016 Torsten Grote
+ *    Quickhome-Widget
+ *    Copyright (C) 2016 Patrick Kanzler
```
grote: 2017 ;)

src/de/grobox/liberario/ui/LocationGpsView.java
```diff
@@ -160,6 +160,7 @@ public void activateGPS() {
 			locationManager.removeUpdates(this);
 			Toast.makeText(getContext(), getContext().getString(R.string.error_no_location_provider), Toast.LENGTH_LONG).show();
 
+			searching = false;
```
grote: Why was this added?

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -533,7 +564,7 @@ private void showMore(boolean animate) {
 
 	private void showLess(boolean animate) {
 		showingMore = false;
-		ui.via.setLocation(null);
+		ui.via.setLocation((Location)null);
```
grote: Is this really necessary?

res/layout/widget_quickhome.xml
patkan: Nope, I will read about this...

src/de/grobox/liberario/ui/LocationGpsView.java
patkan: Otherwise the GpsLocationView "hangs" on searching. It won't search again because searching is true and it stopped searching at the same time. You can reproduce this by doing a normal search with GPS disabled.
I assumed this was unintended behavior. Moreover this allows the search, once started by the appwidget to abort automatically and not hang in the ProcessDialog.

I'm going to add a comment if this stays in the code.

src/de/grobox/liberario/fragments/DirectionsFragment.java
patkan: When I was putting this together it was. I will check if or how this can be removed now.

patkan: I haven't found a real solution for this. I just moved the need for the cast into a method of the LocationView. I don't know if this makes it better.

patkan:
OK, now this is more or less done. Still open or unsure:

* the layout of the widget is still not perfect. I don't know how to do it better and am open for ideas. I am currently researching if maybe a shortcut (not the new Android7.1 ones) would be better.
* Some user experience (like when GPS is disabled) could be better, but I think I found all outright errors.
* rebase to current master (I can do that any time.)
* stash, I assume you stash this into one commit nevertheless, so this is not really needed.

src/de/grobox/liberario/fragments/DirectionsFragment.java
grote: No this doesn't make it better, sorry :P

Maybe if you tell me what the problem is, I can help? Is it just a warning? In any case, if you want to pass `null` as `Location`, you should annotate the parameter of `setLocation()` method as `@Nullable`.

grote:
> the layout of the widget is still not perfect. I don't know how to do it better and am open for ideas.

I could also see if I can improve it further once this got merged.

> I am currently researching if maybe a shortcut (not the new Android7.1 ones) would be better.

Interesting. Let me know what you find out (even if negative)!

> Some user experience (like when GPS is disabled) could be better

What exactly could be better?

> but I think I found all outright errors.

When you declare it ready, I'll do some testing of my own.

> rebase to current master (I can do that any time.)
> stash, I assume you stash this into one commit nevertheless, so this is not really needed.

Both is actually not needed. GitHub does these automatically now when I do a "Squash and merge".

PS. The travis fixes seem to have worked! 👍

src/de/grobox/liberario/fragments/DirectionsFragment.java
patkan: No problem. I will revert my last commit except for the annotations.

The problem is that `setLocation(null)` is ambiguous. It can be both `setLocation(WrapLocation)` or `setLocation(Location)` since they both accept a single argument. Passing `null` to `setLocation(WrapLocation)`, however, does not make any sense. (This method can't even handle null, I should probably fix this.)

grote: How about changing one of the methods to `setWrapLocation()`?

Alternatively, we could introduce a `clearLocation()` method and replace all `setLocation(null)` calls with it. This is tricky however when a variable that can be `null` is passed.

patkan:
> > Some user experience (like when GPS is disabled) could be better

> What exactly could be better?

I just thought of maybe starting the search automatically when GPS gets reenabled. For now the search just aborts. But when I think about that it is maybe the right way to handle this case. (Why should the search start automatically after the toast that GPS is disabled has been shown?)

Considering the shortcuts: I found some old [stackoverflow from 2014](https://stackoverflow.com/questions/25334555/is-it-possible-to-get-the-shortcut-layout-of-launcher-and-use-it-in-our-own-widg?rq=1) that shows how to create launcher shortcuts with custom intents. I still have to try that.

In addition with API25 a new type of shortcuts has been introduced: [App shortcuts](https://developer.android.com/guide/topics/ui/shortcuts.html). I do not know however if it is possible to have "optional" API-coverage. (If I understand right we would have to build especially for API25 for this feature.)

grote:
> I just thought of maybe starting the search automatically when GPS gets reenabled. For now the search just aborts. But when I think about that it is maybe the right way to handle this case. (Why should the search start automatically after the toast that GPS is disabled has been shown?)

No, I think it is fine when the user needs to press search again after enabling GPS.

> Considering the shortcuts: I found some old stackoverflow from 2014 that shows how to create launcher shortcuts with custom intents. I still have to try that.

If I understand this right, it is like a second app in the app drawer. The advantage would be that it would be more visible than the widget and easier to handle (no resizing). The disadvantage is that it would appear for everybody in the app list, if they want it or not. People might then wonder what app is this? Did I install this?

> In addition with API25 a new type of shortcuts has been introduced: App shortcuts. I do not know however if it is possible to have "optional" API-coverage. (If I understand right we would have to build especially for API25 for this feature.)

Yes, I have read about those. If we use them, they would only work for API 25+ devices, so if you want to do that , this would be great as a separate PR and a parallel way to reach the home feature.

patkan:
I would consider this now ready to be reviewed. I consider the Widget not finished, but did not succeed in improving it. So I would be happy if you could have a look at it.

src/de/grobox/liberario/fragments/DirectionsFragment.java
```diff
@@ -443,29 +444,59 @@ private void processIntent() {
...
+				ui.to.setWrapLocation(wto);
+			}
+		} else {
+			to = null;
+		}
+
+		presetFromTo(from, via, to, date);
+	}
+
 	private void presetFromTo(Location from, Location via, Location to, Date date) {
```
grote: Can this method and the one above maybe be merged? Maybe all callers of `presetFromTo()` for `Location` parameters can be changed to `WrapLocation`?

src/de/grobox/liberario/ui/LocationView.java
```diff
@@ -331,11 +331,32 @@ public void setLocation(Location loc, Drawable icon) {
...
 	}
 
-	public void setLocation(Location loc) {
+	public void setLocation(@Nullable Location loc) {
 		Drawable drawable = getDrawableForLocation(getContext(), loc);
 		setLocation(loc, drawable, true);
 	}
 
+	public void setWrapLocation(WrapLocation loc) {
+		if(loc == null) {
```
grote: This parameter is not annotated `@Nullable`. Can it be `null`?

grote:
Oh btw. rebasing this would be nice as it makes testing a lot easier.

grote:
FYI: I am planning a major refactoring for Transportr 2. It would be nice if we could merge this before to avoid merge conflicts.

patkan:
Thanks for the information and sorry for taking so long.
I will try to get that asap ready for merging.

grote:
No need to drop everything and rush. Just giving you a heads up ;)

src/de/grobox/liberario/fragments/DirectionsFragment.java
patkan: I could do that but that would make the method `presetFromTo()` more complex. I will refactor this and then we can see.

patkan:
> Another thing I noticed which wasn't there before is that when you press 'Back' from the trip list, the search is made again. Maybe the intent needs to be cleared when the search is made? It does not seem to happen when I click a recent trip for example. Would be interesting to know why there's a difference. Both should be using intents.

Interesting. I am sorry but I can't seem to reproduce that error. I will search the code if there is any point that might produce this problem.

src/de/grobox/liberario/fragments/DirectionsFragment.java
patkan: In fact I realized that there is no caller of `presetFromTo(Location...` left. So I just moved the code of both methods into each other.

patkan:
I will rebase this now to the current master and force-push into my branch.

grote:
Maybe too late for this PR, but this looks very interesting: http://stackoverflow.com/a/20536046/4856311

Would allow us to set arbitrary searches as launcher shortcuts and would also avoid widget scaling issues.

grote:
> I am sorry but I can't seem to reproduce that error.

I found another way to reproduce this while testing on another phone:

1. Put your phone well indoors, so it won't get a GPS fix immediately. Turn off network based locations.
2. Click the home widget
3. Cancel the dialog that says it is looking for the current location
4. Click the first location input field and select "get location from map"
5. Press back right away
6. Observe how the DirectionFragment behaves as if you just had clicked the home widget

src/de/grobox/liberario/ui/LocationView.java
```diff
@@ -305,6 +305,10 @@ protected LocationAdapter getAdapter() {
 		return (LocationAdapter) ui.location.getAdapter();
 	}
 
+	public boolean getChangingHome() {
```
grote: great fix! style nipick: `isChangingHome()` might be a better name for this method.

patkan:
When I follow your description (I have switched localization to "device only", so use only GPS), I get the following state. However no new search is started. Is this the state that you meant? I am asking myself if I do something "wrong" or where there is a difference.
I still have to debug the intents so maybe I will find something there.
screenshot_2017-01-17-15-21-48

grote:
No that is not what is happening here (on different phones). When getting back to the DirectionFragment (no matter where from) the search is started again exactly as if you had not pressed back, but clicked the home widget.

Maybe you can try on a different phone or an emulator? If you can't reproduce, I can also look into the bug.

patkan:
On which version of Android are you? I could not reproduce until now on API 22. On API 21 in an emulator I seem not to be able to reproduce this consistently. (I was once clicking around and managed to get the from-icon blinking).

grote:
I observed the back-searching issue on API 22 and 16 so far.

AndroidManifest.xml
```diff
@@ -12,6 +12,8 @@
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 
+    <uses-permission android:name="com.android.launcher.permission.INSTALL_SHORTCUT" />
```
grote: Awesome you are trying that! :)

Is this a "dangerous" permission that we need to ask the user to grant us?

Do you want to remove the widget or offer it as an alternative option?

patkan:
Yep, I'm trying it :-) By the way: I could unfortunately not reproduce the error both in emulator and on my smartphone. I am a bit at loss what it could be.

I tried to add the shortcut-creation code but when I execute it, my launcher crashes (I tried it with several launchers on different systems).
The problem seems to be that it can't find the WrapLocation-class in its scope.
I tried to work around this by modifying the intent-handler a bit. @grote what do you think of this fix? I am not 100% with it, but I can't imagine a better way at the moment.


```
01-21 22:07:16.053 20897-20897/? E/AndroidRuntime: FATAL EXCEPTION: main
                                                   Process: com.fairphone.fplauncher3, PID: 20897
                                                   java.lang.RuntimeException: Unable to start receiver com.fairphone.fplauncher3.InstallShortcutReceiver: java.lang.RuntimeException: Parcelable encountered ClassNotFoundException reading a Serializable object (name = de.grobox.liberario.WrapLocation)
                                                       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2616)
                                                       at android.app.ActivityThread.access$1700(ActivityThread.java:151)
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1380)
                                                       at android.os.Handler.dispatchMessage(Handler.java:102)
                                                       at android.os.Looper.loop(Looper.java:135)
                                                       at android.app.ActivityThread.main(ActivityThread.java:5254)
                                                       at java.lang.reflect.Method.invoke(Native Method)
                                                       at java.lang.reflect.Method.invoke(Method.java:372)
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904)
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699)
                                                    Caused by: java.lang.RuntimeException: Parcelable encountered ClassNotFoundException reading a Serializable object (name = de.grobox.liberario.WrapLocation)
                                                       at android.os.Parcel.readSerializable(Parcel.java:2384)
                                                       at android.os.Parcel.readValue(Parcel.java:2203)
                                                       at android.os.Parcel.readArrayMapInternal(Parcel.java:2485)
                                                       at android.os.BaseBundle.unparcel(BaseBundle.java:221)
                                                       at android.os.BaseBundle.keySet(BaseBundle.java:321)
                                                       at android.content.Intent.toUriInner(Intent.java:7417)
                                                       at android.content.Intent.toUriFragment(Intent.java:7368)
                                                       at android.content.Intent.toUri(Intent.java:7359)
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.addToInstallQueue(InstallShortcutReceiver.java:129)
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.onReceive(InstallShortcutReceiver.java:312)
                                                       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2609)
                                                       at android.app.ActivityThread.access$1700(ActivityThread.java:151) 
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1380) 
                                                       at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                       at android.os.Looper.loop(Looper.java:135) 
                                                       at android.app.ActivityThread.main(ActivityThread.java:5254) 
                                                       at java.lang.reflect.Method.invoke(Native Method) 
                                                       at java.lang.reflect.Method.invoke(Method.java:372) 
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904) 
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699) 
                                                    Caused by: java.lang.ClassNotFoundException: de.grobox.liberario.WrapLocation
                                                       at java.lang.Class.classForName(Native Method)
                                                       at java.lang.Class.forName(Class.java:309)
                                                       at java.io.ObjectInputStream.resolveClass(ObjectInputStream.java:2263)
                                                       at android.os.Parcel$2.resolveClass(Parcel.java:2375)
                                                       at java.io.ObjectInputStream.readNewClassDesc(ObjectInputStream.java:1641)
                                                       at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:657)
                                                       at java.io.ObjectInputStream.readNewObject(ObjectInputStream.java:1782)
                                                       at java.io.ObjectInputStream.readNonPrimitiveContent(ObjectInputStream.java:761)
                                                       at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1983)
                                                       at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1940)
                                                       at android.os.Parcel.readSerializable(Parcel.java:2378)
                                                       at android.os.Parcel.readValue(Parcel.java:2203) 
                                                       at android.os.Parcel.readArrayMapInternal(Parcel.java:2485) 
                                                       at android.os.BaseBundle.unparcel(BaseBundle.java:221) 
                                                       at android.os.BaseBundle.keySet(BaseBundle.java:321) 
                                                       at android.content.Intent.toUriInner(Intent.java:7417) 
                                                       at android.content.Intent.toUriFragment(Intent.java:7368) 
                                                       at android.content.Intent.toUri(Intent.java:7359) 
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.addToInstallQueue(InstallShortcutReceiver.java:129) 
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.onReceive(InstallShortcutReceiver.java:312) 
                                                       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2609) 
                                                       at android.app.ActivityThread.access$1700(ActivityThread.java:151) 
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1380) 
                                                       at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                       at android.os.Looper.loop(Looper.java:135) 
                                                       at android.app.ActivityThread.main(ActivityThread.java:5254) 
                                                       at java.lang.reflect.Method.invoke(Native Method) 
                                                       at java.lang.reflect.Method.invoke(Method.java:372) 
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904) 
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699) 
                                                    Caused by: java.lang.ClassNotFoundException: Didn't find class "de.grobox.liberario.WrapLocation" on path: DexPathList[[zip file "/system/priv-app/FairphoneLauncher3/FairphoneLauncher3.apk"],nativeLibraryDirectories=[/vendor/lib, /system/lib]]
                                                       at dalvik.system.BaseDexClassLoader.findClass(BaseDexClassLoader.java:56)
                                                       at java.lang.ClassLoader.loadClass(ClassLoader.java:511)
                                                       at java.lang.ClassLoader.loadClass(ClassLoader.java:469)
                                                       at java.lang.Class.classForName(Native Method) 
                                                       at java.lang.Class.forName(Class.java:309) 
                                                       at java.io.ObjectInputStream.resolveClass(ObjectInputStream.java:2263) 
                                                       at android.os.Parcel$2.resolveClass(Parcel.java:2375) 
                                                       at java.io.ObjectInputStream.readNewClassDesc(ObjectInputStream.java:1641) 
                                                       at java.io.ObjectInputStream.readClassDesc(ObjectInputStream.java:657) 
                                                       at java.io.ObjectInputStream.readNewObject(ObjectInputStream.java:1782) 
                                                       at java.io.ObjectInputStream.readNonPrimitiveContent(ObjectInputStream.java:761) 
                                                       at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1983) 
                                                       at java.io.ObjectInputStream.readObject(ObjectInputStream.java:1940) 
                                                       at android.os.Parcel.readSerializable(Parcel.java:2378) 
                                                       at android.os.Parcel.readValue(Parcel.java:2203) 
                                                       at android.os.Parcel.readArrayMapInternal(Parcel.java:2485) 
                                                       at android.os.BaseBundle.unparcel(BaseBundle.java:221) 
                                                       at android.os.BaseBundle.keySet(BaseBundle.java:321) 
                                                       at android.content.Intent.toUriInner(Intent.java:7417) 
                                                       at android.content.Intent.toUriFragment(Intent.java:7368) 
                                                       at android.content.Intent.toUri(Intent.java:7359) 
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.addToInstallQueue(InstallShortcutReceiver.java:129) 
                                                       at com.fairphone.fplauncher3.InstallShortcutReceiver.onReceive(InstallShortcutReceiver.java:312) 
                                                       at android.app.ActivityThread.handleReceiver(ActivityThread.java:2609) 
                                                       at android.app.ActivityThread.access$1700(ActivityThread.java:151) 
                                                       at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1380) 
                                                       at android.os.Handler.dispatchMessage(Handler.java:102) 
                                                       at android.os.Looper.loop(Looper.java:135) 
                                                       at android.app.ActivityThread.main(ActivityThread.java:5254) 
                                                       at java.lang.reflect.Method.invoke(Native Method) 
                                                       at java.lang.reflect.Method.invoke(Method.java:372) 
                                                       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:904) 
                                                       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:699) 
                                                   	Suppressed: java.lang.ClassNotFoundException: de.grobox.liberario.WrapLocation
                                                       at java.lang.Class.classForName(Native Method)
                                                       at java.lang.BootClassLoader.findClass(ClassLoader.java:781)
                                                       at java.lang.BootClassLoader.loadClass(ClassLoader.java:841)
                                                       at java.lang.ClassLoader.loadClass(ClassLoader.java:504)
                                                       		... 31 more
                                                    Caused by: java.lang.NoClassDefFoundError: Class not found using the boot class loader; no stack available
```

AndroidManifest.xml
patkan: As far as I see it, this should work without user interaction, but I will just test later by "manually" installing the apk.
I would offer it as an alternative to the widget, but if you think the user experience is better with just one option this would also be ok for me.

patkan: Found it:
https://developer.android.com/reference/android/Manifest.permission.html#INSTALL_SHORTCUT
INSTALL_SHORTCUT has protection level normal. When I install it is listed as a permission but it should not lead to a prompt for this permission.

grote: Nice! But see `Added in API level 19`. So you can only use this on this API level and need to check before on what API you are.

patkan:
I added the creation of a shortcut to the settings menu. This is not the perfect place, but I don't see a better option. What do you think, @grote ?
Is there a possibility to hide the preference on lower APIs? I haven't found one.

As always I am open for improvements and suggestions.

grote:
I think the settings menu is a good place for adding this shortcut. In Transportr 2 there will be a better place, but for now it is good as it is.

You should be able to hide the preference, after the preferences got inflated. Find the preference there, and set in invisible/hidden when the API level is too low. You already do `findPreference()`. It should work right after that. Something like `preferenceScreen.removePreference(home)`;

Regarding the WrapLocation stacktrace, I think that is expected when the launcher doesn't know about this class. You solution for this looks good to me. Only nitpick, static final Strings should be all uppercase `s/taskBringMeHome/TASK_BRING_ME_HOME/`

Is there anything else outstanding in this PR, or should I give it another test?

src/de/grobox/liberario/fragments/SettingsFragment.java
```diff
@@ -92,30 +93,30 @@ public boolean onPreferenceClick(Preference preference) {
...
-					addIntent.putExtra(Intent.EXTRA_SHORTCUT_ICON_RESOURCE, Intent.ShortcutIconResource.fromContext(getContext(), R.drawable.ic_quickhome_widget));
-					addIntent.putExtra("duplicate", false);
-					addIntent.setAction("com.android.launcher.action.INSTALL_SHORTCUT");
-					getContext().sendBroadcast(addIntent);
 					Toast.makeText(getContext(), getString(R.string.toast_create_quickhome_shortcut), Toast.LENGTH_SHORT).show();
 				} else {
-					Toast.makeText(getContext(), getString(R.string.error_cannot_create_quickhome_shortcut), Toast.LENGTH_SHORT).show();
+					// the INSTALL_SHORTCUT permission is part of the launcher below API 19 and
+					// therefore might not be supported
+					Toast.makeText(getContext(), getString(R.string.toast_create_quickhome_old_API), Toast.LENGTH_SHORT).show();
```
grote: I think just hiding the preference on unsupported platforms is the better approach. You don't need an error Toast then.

patkan: The "problem" is that this seems to be supported since Android 1.x but it is only a system-wide permission since API 19.
I tested this on API 18 and it just works fine. (Formerly this permission was part of the launcher)

patkan:
This should be ready now. The following points are still open:
* the bug(s) that you described (I can't reproduce them and don't understand why)
* the widget is below perfect
* missing strategy for below API 19 for the shortcut-adding-code

I am still working on the last point.

src/de/grobox/liberario/fragments/SettingsFragment.java
patkan: I liked it more this way, because the user can try if it works (if not, this is just sending an Intent).
If you prefer I can also remove the permission below API 19.

grote:
I tested this PR once more.

The launcher shortcut looks very nice (better than the widget). I was personally somewhat confused by the Toast text. Maybe you should show the launcher right away (instead of the Toast) after adding the shortcut. Mine showed a nice animation after pressing the home button.

Creating a shortcut worked on my API 16 device. Transportr 2 will require 15. So I'd say get rid of the toasts and just [show the home screen](http://stackoverflow.com/a/3725042/4856311), then wait for bug reports of people where it didn't work ;)

Unfortunately, the back-re-search problem is still there. I would send you a video, but don't want to disclose my location ;)
After debugging this a bit more it turns out the bug has to do with the useful Android development option `Don't keep activities`. Even though the intent gets cleared in `DirectionsFragment#processIntent()`, the recreated activity still has it and searches again. I am not sure why that is. Maybe it has to do with the intent flags you have chosen. Any reason for those?

(Also, when I cancel the location search and change the from location, the icon keeps blinking and the chosen location is eventually replaced by the GPS coordinates. Looks like the GPS search isn't canceled anymore when I choose another location. But this doesn't seem to be your bug, but already present in the latest release, so please don't worry about it.)

Great work so far! :) Let's resolve the last issues and then get this merged!

res/layout/widget_quickhome.xml
```diff
@@ -0,0 +1,17 @@
...
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
+             xmlns:app="http://schemas.android.com/apk/res-auto"
+             xmlns:tools="http://schemas.android.com/tools"
+             android:layout_width="match_parent"
+             android:layout_height="match_parent"
+             android:id="@+id/widget_quickhome_frame">
+
+    <ImageView
+        android:layout_width="64dp"
+        android:layout_height="64dp"
```
grote: Have you tried `wrap_content` here? Maybe then the widget will look nicer?

grote:
> Even though the intent gets cleared in `DirectionsFragment#processIntent()`, the recreated activity still has it and searches again. I am not sure why that is. 

I think it is because when clicking the home shortcut, the activity is created from scratch with this intent. So when it is recreated, the same intent is used. Resetting it in the old activity doesn't help then. The other callers for this intent fire when the activity is already started, so there the recreation is no problem.

While activities can always be killed by Android, if we don't find a solution quickly, maybe we just ignore this issue for now? In Transportr 2, the situation will be different and *maybe* no issue.

patkan:
>I think it is because when clicking the home shortcut, the activity is created from scratch with this intent. So when it is recreated, the same intent is used. Resetting it in the old activity doesn't help then. The other callers for this intent fire when the activity is already started, so there the recreation is no problem.

This theory sounds good. The flags that I have set are for making sure that the search is executed. Since Android does not look on the Extras, it can't distinguish a "normal" Intent from a "Get Me Home" Intent. That's why it did not create new activities sometimes. (The activity was then prepopulated with an old search.) If I remember correctly, in the case of a launcher shortcut those flags are just for documentation reasons; they will be added nevertheless to the Intent.

With the developer setting that you showed me, I can consistently reproduce your bug-report. However, I could not find a solution to this problem. Maybe we could implement something like a proxy-activity that is put somewhere in the stack so that when we press back we do not return to the activity that has been created with our Intent. I think this would, however, take quite some refactoring.
So for now I would really let it be like this and then see how it turns out in Transportr 2. If the problem persists I'll be happy to think of a solution.

Concerning the app widget:
I thought about it. The only benefit is that it is resizable (in theory, the committed version was not able to do this). All the other stuff is handled better by the shortcut. (You can even place it in folders,) I tried to get the layout nice but I was not able to make it look good in more than one launcher. (I had problems with getting the off-centering right and I had problems with getting the size right.) Therefore, I have removed the app widget. Should it be revealed that the shortcut has any problems we can still readd the widget-code.

So as far as I am concerned this is ready to go. I would like to thank you very much for the great support when creating this. :-)

grote:
Thanks for your work and patience @patkan! :)

I am going to merge this now, but there's one small thing that could maybe be done in another PR: If we have a way to find out whether the shortcut was already installed, we should either hide the preference if installed or at least don't show the home screen and let the user know that the shortcut is already there.

patkan:
> I am going to merge this now, but there's one small thing that could maybe be done in another PR: If we have a way to find out whether the shortcut was already installed, we should either hide the preference if installed or at least don't show the home screen and let the user know that the shortcut is already there.

I think there is no way to check for a shortcut, but I will search. I will open a PR if I have something.

PR #266, opened 2016-11-29
SNCB (Belgium)

editicalu:
First of all, this is a minor issue. No big changes are involved, as you can see at the code.

I live in Belgium and like the app a lot, but the Belgium implementation wasn't clear. The used service (SNCB) provides every public transport you could use in Belgium. However, the description said "trains in Belgium". That's why I've updated this. The networks provided are the bus, metro and tram connections around the country.

Besides of that, the name of the SNCB isn't correct. Because Belgium is a country with 3 official languages, the official name of the railway service is SNCB-NMBS. SNCB (as provided) is only the French name.

Finally, I've updated the logo to a svg version.


grote:
Thanks for the PR, looks great!

There's only one issue: Did you try if your new vector logo is taken? It looks like the PR is missing the Android XML vector drawable that can be created from the SVG.

res/values/strings.xml
```diff
@@ -206,7 +206,8 @@ there are any).</string>
     <string name="np_desc_wien">Vienna</string>
 
     <string name="np_region_belgium">Belgium</string>
-    <string name="np_desc_sncb">Trains in Belgium</string>
+    <string name="np_desc_sncb">Public Transport in Belgium, by the SNCB-NMBS</string>
+    <string name="np_desc_sncb_networks" translatable="false">SNCB-NMBS, De Lijn, TEC, MIVB-STIB</string>
```
grote: Doesn't look like this string is actually used anyway, yet.

editicalu:
I've tested it now and swapped the logo to the flag, because otherwise, people wouldn't be able to see anything on the logo.

src/de/grobox/liberario/TransportNetworks.java
```diff
@@ -1,4 +1,4 @@
-/*    Transportr
```
grote: Why did you remove that?

src/de/grobox/liberario/TransportNetworks.java
```diff
@@ -297,7 +297,8 @@
 		region = region(R.string.np_region_belgium, "\uD83C\uDDE7\uD83C\uDDEA");
 
 		list.add(new TransportNetwork(context, NetworkId.SNCB)
-				         .setDescription(getString(R.string.np_desc_sncb))
+		                 .setName(getString(R.string.np_region_belgium))
+				         .setDescription(getString(R.string.np_desc_sncb), getString(R.string.np_desc_sncb_networks))
```
grote: Style nitpick: here are some superfluous spaces.

grote:
Thanks for the update! The PR now contains two SVG files, but only one of them is being used/needed.

Why are you not using the SCNF logo in the app now? I would reserve the Belgium flag for when the data comes from Navitia that aggregates various GTFS data. This data here is actually coming from SCNF, so I would prefer using their logo. If there's an issue with converting the logo to Android XML, let me know.

editicalu:
1) I created the svg with the logo's first, but I thought this wouldn't
be readable on a smaller screen.
2) I've used the flag because it contains every form of national public
transport available, not only the trains.
3) The logo's-svg couldn't be converted easily to XML.

PR #267, opened 2016-12-05
Update description Spain

tenminutemail:
I updated the description and removed Spain form it, as the renfe data available only works in Basque Country


grote:
Awesome, thanks a lot! :)

PR #269, opened 2016-12-06
Update SBB info

tenminutemail:
Idea: change "Trains in Switzerland" to "Switzerland", as the SBB data contains basically all public transport info for Switzerland. The OEBB data is also described as Austria (not "Trains in Austria") because there is more information available.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -344,7 +344,7 @@ there are any).</string>
     <string name="np_region_switzerland">Switzerland</string>
     <!-- SBB is german/international, CFF is french and FFS is italian! So this IS translatable. -->
     <string name="np_name_sbb" translatable="true">SBB</string>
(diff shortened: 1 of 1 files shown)


tenminutemail:
I also have an ZVV logo, but the upload doens't work, so I'll upload it here (as we're talking about Switzerland). There is an SVG available at Wikipedia.

network_zvv_logo

grote:
@tenminutemail Thanks for the PR! :)

@patkan I think you were involved with the current description. Are you fine with that change?

Is it correct that the SBB provider does return for results for for then trains and does so in entire Switzerland?

patkan:
I am fine with this change.
I think I missed the fact that it does not only list trains.
I checked it in the area around Lausanne and there it returned also busses. I have however no real experience with this provider and thus can't tell more.

PR #275, opened 2016-12-11
New Logos and Change name provider Liechtenstein

tenminutemail:
Hi,

Two more logos: one for SEPTA, US and another one for Liechtenstein, VVV.

I've changed the name from the Liechtenstein provider from VAO to VVV, because according to Wikipedia (https://de.wikipedia.org/wiki/Liechtenstein#Busverkehr & https://de.wikipedia.org/wiki/Verkehrsverbund_Vorarlberg#Verkehrsunternehmen_im_VVV) this is the only transport authority that operates in Liechtenstein (& Austria). I think it was called VAO because the data comes from here: http://www.verkehrsauskunft.at/

Idea: What about including this VVV network also to the Austrian networks, as Vorarlberg is a federal state of Austria? So that it would be available for Liechtenstein and Austria


res/values/strings.xml
```diff
@@ -311,6 +311,7 @@ there are any).</string>
     <string name="np_desc_it_networks" translatable="false">ATM, GTT, AMAT, ACTV</string>
 
     <string name="np_region_liechtenstein">Liechtenstein</string>
+    <string name="np_name_vmobil">VVV</string>
```
grote: Did you test if this is actually doing anything? I suspect it is not.

grote:
Thanks for the new logos! Usually, the name matches where the name comes from. Maybe @sirnone has more insight into why the Liechtenstein provider might be named VAO (and still works) and not VVV.

There is already a VVVProvider btw.: https://github.com/schildbach/public-transport-enabler/blob/931dafb628c23733c162bd23972c97ef1a72e642/enabler/src/de/schildbach/pte/VvvProvider.java
Verkehrsverbund Vogtland

> Idea: What about including this VVV network also to the Austrian networks, as Vorarlberg is a federal state of Austria? So that it would be available for Liechtenstein and Austria

Good idea.

res/values/strings.xml
tenminutemail: No, I didn't test it. I suppose it's just renaming the "VAO" to "VVV", similar to what happens with db here:
```
    <string name="np_name_db" translatable="false">Deutsche Bahn</string>
    <string name="np_desc_db">Most of Germany</string>
```

tenminutemail: Okay, so how can the network be renamed?

tenminutemail:
About including this network to Austria as well, what needs to be changed? The strings.xml?

tenminutemail:
@grote I just removed the change I made to the strings.xml. 

Anyway, I'm still intrigued on how to change the name from VAO to VVV, in the "pick a transportation network" list.

grote:
Merged, thanks!

I'll change the name.

PR #278, opened 2016-12-14
Add support for Navitia fr-se region (south-east of France)

dmaulat:
I mainly tested the south-east region in the surroundings of Grenoble (TAG network). Has worked like a charm for 2 days straight :-)
Support for Lyon (2nd biggest french city) is lacking from Navitia, however.

### DIFF
--- a/res/drawable/network_francesoutheast_logo.xml
+++ b/res/drawable/network_francesoutheast_logo.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="64dp"
+        android:height="64dp"
+        android:viewportHeight="64"
+        android:viewportWidth="64">
+
+    <path
+        android:fillColor="#ed2939"
+        android:pathData="M42.66,0 L64,0 L64,64 L42.66,64 Z"/>
+    <path
+        android:fillColor="#ffffff"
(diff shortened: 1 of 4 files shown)


grote:
Hey, thanks a lot for your PR! I didn't know South-east was in PTE already.

Will look at it in more detail later. Only so much now: Please do different PRs for different features and do not update the translation in the code. This should only be done via transifex.

dmaulat:
Alright, sorry about that. I can revert my code about the translation if need be.

dmaulat:
And btw, Anthony Chaput seems to have added support in PTE for both South-West and South-East in back in August. I only had to rely on it.

grote:
> Alright, sorry about that. I can revert my code about the translation if need be.

Yes, please don't touch the translations in code. The issue is that either your translations will get overridden by the ones from Transifex or the other way around in any case, people's work will be lost which is not good. Maybe there should be a note about this in https://github.com/grote/Transportr/blob/master/CONTRIBUTING.md if people are reading that ;)

dmaulat:
> Yes, please don't touch the translations in code.

I rewrote my commit to be more atomic, as you wished. I will update the translations via Transifex.

> Maybe there should be a note about this in https://github.com/grote/Transportr/blob/master/CONTRIBUTING.md if people are reading that ;)

Hi, yes, I read this file beforehand. I think you are right, it could be useful to provide specific instructions for translating. It's the first time I contribute code and I didn't think that far ahead, so other beginners may be confused as well.

grote:
Thanks a lot for making the changes. Looks very good now!

Issue #286, opened 2016-12-29
Language of name of places


Airon90:
Are name of cities and regions written in English or in the local language?
Some places are written in English (Upper Austria, Saxony) while other are written in local language (Corse, Roma, Milano, Venezia).
Let's decide a standard and adapt all names.

grote:
in `values/strings.xml` the names should be in English.

Airon90:
Ok, I'm assigning this bug ;)

grote:
Fixed by  684a7d1

Airon90:
I didn't check all names as yesterday I didn't have much spare time. That's why I added (1/2). Today I'll check all the remaining places ;)

PR #287, opened 2016-12-29
Set name of places in English (1/2)

Airon90:
Fixing #286,
Fixed wrong year (next week it should be changed again lol)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -347,7 +347,7 @@ there are any).</string>
     <string name="np_name_sbb" translatable="true">SBB</string>
     <string name="np_desc_sbb">Switzerland</string>
     <string name="np_desc_vbl">Lucerne</string>
-    <string name="np_desc_zvv">Zürich</string>
+    <string name="np_desc_zvv">Zurich</string>
 
     <string name="np_region_uae">United Arab Emirates</string>
     <string name="np_name_dub">Dubai</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -48,7 +48,7 @@ there are any).</string>
     <string name="tab_devs">Developers</string>
     <string name="tab_about">About</string>
 
-    <string name="about_top">© 2013 &#8211; 2015 Torsten Grote\nFree Software License: GPLv3</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -346,8 +346,8 @@ there are any).</string>
     <!-- SBB is german/international, CFF is french and FFS is italian! So this IS translatable. -->
     <string name="np_name_sbb" translatable="true">SBB</string>
     <string name="np_desc_sbb">Switzerland</string>
-    <string name="np_desc_vbl">Luzern</string>
-    <string name="np_desc_zvv">Zurich</string>
+    <string name="np_desc_vbl">Lucerne</string>
+    <string name="np_desc_zvv">Zürich</string>
```
grote: Are you sure Zürich is English?

res/values/strings.xml
```diff
@@ -48,7 +48,7 @@ there are any).</string>
     <string name="tab_devs">Developers</string>
     <string name="tab_about">About</string>
 
-    <string name="about_top">© 2013 &#8211; 2015 Torsten Grote\nFree Software License: GPLv3</string>
+    <string name="about_top">© 2013 &#8211; 2016 Torsten Grote\nFree Software License: GPLv3</string>
```
grote: Please change it to 2017 right away.

res/values/strings.xml
Airon90: In the [Wikipedia page](https://en.wikipedia.org/wiki/Z%C3%BCrich) it is used "Zürich" even if it is written that even Zurich is used

res/values/strings.xml
Airon90: Done ;)

res/values/strings.xml
grote: Thanks! However, I still prefer to leave it as "Zurich" as this is also correct and the main reason: Every time you change a source string *all* translations for this string have to be redone. I don't think this change justifies creating unnecessary work for all translators.

Airon90: Done ;)

grote:
Thanks again! :)

PR #289, opened 2017-01-03
Open "Pick a network" view when tabbing on the current network

### LINKED ISSUE #282: Tabbing on current network should go to network picker view
IMHO it was more intuitive to go to the "pick a transportation network" view after tabbing on the icon of the current network in the side bar instead of closing the side bar.

sedrubal:
Fixes #282 

If you think, this is not intuitive just close this PR :wink:

### DIFF
--- a/src/de/grobox/liberario/activities/MainActivity.java
+++ b/src/de/grobox/liberario/activities/MainActivity.java
@@ -114,7 +114,11 @@ public class MainActivity extends TransportrActivity implements FragmentManager.
              .withOnAccountHeaderListener(new AccountHeader.OnAccountHeaderListener() {
 	             @Override
 	             public boolean onProfileChanged(View view, IProfile profile, boolean currentProfile) {
-		             if(!currentProfile && profile != null && profile instanceof ProfileDrawerItem) {
+		             if(currentProfile) {
+			             Intent intent = new Intent(getContext(), PickNetworkProviderActivity.class);
+			             ActivityOptionsCompat options = ActivityOptionsCompat.makeScaleUpAnimation(getCurrentFocus(), 0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


src/de/grobox/liberario/activities/MainActivity.java
```diff
@@ -114,7 +114,11 @@ public void onCreate(Bundle savedInstanceState) {
              .withOnAccountHeaderListener(new AccountHeader.OnAccountHeaderListener() {
 	             @Override
 	             public boolean onProfileChanged(View view, IProfile profile, boolean currentProfile) {
-		             if(!currentProfile && profile != null && profile instanceof ProfileDrawerItem) {
+		             if(currentProfile) {
+			             Intent intent = new Intent(getContext(), PickNetworkProviderActivity.class);
+			             ActivityOptionsCompat options = ActivityOptionsCompat.makeScaleUpAnimation(getCurrentFocus(), 0, 0, 0, 0);
+			             ActivityCompat.startActivityForResult(MainActivity.this, intent, CHANGED_NETWORK_PROVIDER, options.toBundle());
+		             } else if(profile != null && profile instanceof ProfileDrawerItem) {
```
grote: Looks like you duplicated this code from elsewhere. How about creating a dedicated method for that code and call it from those two places?

Have you tested this? I ask because you seem to return `false` which means that the event was not consumed. Do you know what effect that has? Will this event then picked up by other code? What happens if you return `true` instead?

sedrubal: Thanks for the review. You're right. I added a `openPickNetworkProviderActivity()` method. I hope this is ok.

I also return `true` now. I don't see any change but I don't know what it means to consume the event (I'm an android newby 🙈 )

grote:
Thanks, looks good now!

PR #290, opened 2017-01-09
Try to fix failing Travis builds

patkan:
I noticed that travis sometimes fails with a killed process.
I found some hints that this might not be the case when `sudo:true` is set. As far as I could test it (I started the build during different daytimes) this seems to solve the sporadic failures.

Moreover I enabled caching of dependencies and build for both `oraclejdk7`, `oraclejdk8` and `openjdk7`.
Those last two are just optional, I can also remove them.
The downturn of `sudo:true` is that it takes a bit longer to start up the machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: android
 before_script: ./pre-build.sh && sed -i "s#com.android.tools.build:gradle:.*'#com.android.tools.build:gradle:1.5.0'#" build.gradle
 jdk: openjdk7
+before_cache:
+    - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+    - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+cache:
+    directories:
+        - $HOME/.gradle/caches/
 android:
   components:
     - tools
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,16 @@
 language: android
+sudo: true # travis-ci/travis-ci#5582 implies this could help against killed build-jobs
 before_script: ./pre-build.sh && sed -i "s#com.android.tools.build:gradle:.*'#com.android.tools.build:gradle:1.5.0'#" build.gradle
-jdk: openjdk7
+jdk:
+    - openjdk7
+    - oraclejdk7
+    - oraclejdk8
```
grote: Since this application is Free Software, I would prefer not building it with non-free software here. What is the advantage of adding the proprietary JDK here?

.travis.yml
```diff
@@ -1,6 +1,16 @@
 language: android
+sudo: true # travis-ci/travis-ci#5582 implies this could help against killed build-jobs
```
grote: travis-ci/travis-ci#5582 is not really conclusive, but we can try this. However, if builds still fail, let's revert this.

.travis.yml
patkan: You are right. I thought of maybe checking compatibility with different build setups, but since this triples the tasks execution time in the worst case I will remove this.

.travis.yml
patkan: That was also my intention. I just couldn't test it thoroughly, so time will tell if this is a solution. (An alternative would be trying to reduce memory consumption of gradle IF this is the problem.)

patkan:
I think this should be ready now.

Issue #231, opened 2016-08-30
Display ticket price


sedrubal:
Öffi can show the price for single trip tickets for the trips. (At least for trips in VGN). This would be an useful information, I think.

grote:
Price would be easy to add. The question is just where to show it. UI mockups are welcome.

sedrubal:
That's great :tada: what do you think about adding it to the image at the top:

![transportr-price](https://cloud.githubusercontent.com/assets/5571650/18172005/3c548f46-7064-11e6-9ef1-c2367da42180.png)

grote:
That's a great start. I'd need to check what data is actually available in the API, but I think it has prices per leg (part of the trip). So it might also happen that you have a price for one leg, but not for another. How would this be handled in the above mockup?

Also, I'm concerned that adding one more line to the header causes problems with readability on the background graphics. We want to have specific background graphics at some point (in this example maybe actually showing a subway in Nurnberg) and then all graphics need to be very dark everywhere. This one has a shim added, but it doesn't go so high into the photo.

sedrubal:
I see... So maybe this is better?

![transportr-price2](https://cloud.githubusercontent.com/assets/5571650/18173069/7bbf7fde-7068-11e6-9808-8534a58794e4.png)

But then there is very much information in this view...

grote:
Could indeed be a bit much. Maybe there could just be an info icon somewhere that opens a dialog with additional information when clicked? This could include the red messages and if there is a red message, the icon could be red or do you think this would not be clear enough for people that there is important information for that trip?

sedrubal:
I like this idea :smile:

Airon90:
I just want to say that sometimes with just one ticket you can use more vehicles: in Milan with one ticket you can catch buses and trams if you travel within 90 minutes. So you don't have to pay another ticket for the other rides.
When you play with APIs check this out in order to know if the information available is correct. It doesn't have sense to show wrong information.

full-duplex:
It looks nice, but this only works for providers that return fares per leg, and not for providers that return the fares for the whole trip, per network, etc. (I think the current Fare implementation is a little lacking anyway, because it doesn't allow reduced prices other than child/adult or 1st/2nd class)

Why not add a card below the trip with all fare info?

grote:
@full-duplex The idea changed half-way through this ticket. The latest one was to add an info icon that opens a `DialogFragment` that shows the message and fares (if any).

full-duplex:
[...]

PR #291, opened 2017-01-09
Use all available width for leg message

full-duplex:
I don't know if this is the best way to fix this, but it seems to work without breaking anything.

Before:
 

After:

### DIFF
--- a/res/layout/leg_details.xml
+++ b/res/layout/leg_details.xml
@@ -60,8 +60,7 @@
         android:textSize="12sp"
         tools:text="This is an important message for this trip. It may contain a lot of text."
         android:layout_below="@+id/lineView"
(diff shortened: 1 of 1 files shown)


grote:
Hey thanks for this PR! You are on to something here, but I am worried that there is no leg duration in your last (right) screenshot.

full-duplex:
The leg duration is hidden by default (like the leg above it) and only shown when the intermediate stops are shown. Since this leg has no intermediate stops, it only shows/hides the duration of the whole leg on press.

To illustrate:

grote:
Awesome, thanks! :)

Would be nice to know what you think about this idea: https://github.com/grote/Transportr/issues/231#issuecomment-244157003

PR #297, opened 2017-01-23
do not show homescreen after first shortcut installation

patkan:
We cannot now which shortcuts are on the users homescreen. We can just generate intents for install and uninstall.
So the only other option for the case that the shortcut is already installed is storing whether we already did install it. If the user however uninstalls the shortcut we won't take notice. So we can only disable the switching the homescreen.

### DIFF
--- a/src/de/grobox/liberario/Preferences.java
+++ b/src/de/grobox/liberario/Preferences.java
@@ -41,6 +41,7 @@ public class Preferences {
 	private final static String OPTIMIZE = "pref_key_optimize";
 	private final static String EXIT_ON_BACK = "pref_key_exit_app_on_back_press";
 	private final static String SELECTED_PRODUCTS = "_selected_products";
+	public final static String SHORTCUT_QUICKHOME_INSTALLED = "pref_shortcut_quickhome_installed";
 
 	private static String getNetwork(Context context, int i) {
 		SharedPreferences settings = context.getSharedPreferences(PREFS, Context.MODE_PRIVATE);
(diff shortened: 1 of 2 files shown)


grote:
Thanks for this follow up! :)

I was hoping we somehow can get hold if the intent result, like here https://android.googlesource.com/platform/packages/apps/Launcher2/+/684476f9a4fb374d026f17487cbf51c391223cb0/src/com/android/launcher2/InstallShortcutReceiver.java#212 to decide whether it was added or not. If that is indeed not possible, the preference solution sounds good.

patkan:
You're welcome. :)
I will search if it is possible to get the return value!

patkan:
As far as I understand the code of this launcher this information never gets out.
See https://android.googlesource.com/platform/packages/apps/Launcher2/+/684476f9a4fb374d026f17487cbf51c391223cb0/src/com/android/launcher2/InstallShortcutReceiver.java#153
I think it is just used to generate those toasts. So I think we can't access that information.

grote:
Thanks for looking into this and thanks for coming up with a work-around for it.

I am not completely sure it is necessary to do this. What do you think? Could we just leave it as is?

Also, I want to move all non-ui tasks (such as retrieving preferences) away from the UI thread which would make this slightly more complicated.

patkan:
> I am not completely sure it is necessary to do this. What do you think? Could we just leave it as is?

Me neither, this might look to the user just like nothing is happening. With the switching to the homescreen the user will at least see that their press has been registered. So as far as I am concerned we could just close this PR. Maybe we just wait for people reporting if there is anything unexpected.

Connected: Do we want to make it possible to create multiple shortcuts? Like it is implemented now only one shortcut can be created (on launchers that do honor the "duplicate"-extra).

grote:
> So as far as I am concerned we could just close this PR.

Fine with me!

> Maybe we just wait for people reporting if there is anything unexpected.

We can always reopen it or make a new one.

> Do we want to make it possible to create multiple shortcuts? Like it is implemented now only one shortcut can be created (on launchers that do honor the "duplicate"-extra).

There should be no duplicated shortcuts, hence the "duplicate"-extra should stay as it is.

patkan:
OK, I'll close this now. :-)

PR #298, opened 2017-01-25
add quickhome shortcut for Android 7.1

patkan:
In relevance to #71 
This adds a "Get Me Home"-shortcut for Android 7.1 Launcher.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -187,6 +187,7 @@ there are any).</string>
     <string name="changelog_show_full">More…</string>
 
     <string name="widget_name_quickhome">Get Me Home</string>
+    <string name="shortcut_quickhome_long_label">Start search for home.</string>
 
     <string name="np_region_australia">Australia</string>
     <string name="np_name_sydney" translatable="false">Transport NSW</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -187,7 +187,6 @@ there are any).</string>
     <string name="changelog_show_full">More…</string>
 
     <string name="widget_name_quickhome">Get Me Home</string>
-    <string name="shortcut_quickhome_long_label">Start search for home.</string>
 
     <string name="np_region_australia">Australia</string>
     <string name="np_name_sydney" translatable="false">Transport NSW</string>
--- a/src/de/grobox/liberario/activities/MainActivity.java
+++ b/src/de/grobox/liberario/activities/MainActivity.java
@@ -18,10 +18,15 @@
 package de.grobox.liberario.activities;
 
 import android.annotation.SuppressLint;
+import android.annotation.TargetApi;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.content.pm.ShortcutInfo;
+import android.content.pm.ShortcutManager;
 import android.graphics.drawable.Drawable;
+import android.graphics.drawable.Icon;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.StrictMode;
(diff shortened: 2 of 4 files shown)


res/values/strings.xml
```diff
@@ -187,6 +187,7 @@ there are any).</string>
     <string name="changelog_show_full">More…</string>
 
     <string name="widget_name_quickhome">Get Me Home</string>
+    <string name="shortcut_quickhome_long_label">Start search for home.</string>
```
grote: This string sounds a bit strange. Why didn't you just re-use the one above? Is there usually a dot at the end of these shortcuts?

src/de/grobox/liberario/fragments/SettingsFragment.java
```diff
@@ -95,13 +95,8 @@ public boolean onPreferenceClick(Preference preference) {
...
 			public boolean onPreferenceClick(Preference preference) {
 				// create launcher shortcut
-				Intent shortcutIntent = new Intent(DirectionsFragment.TAG, Uri.EMPTY, getContext(), MainActivity.class);
-				shortcutIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-				shortcutIntent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
-				shortcutIntent.putExtra("special", DirectionsFragment.TASK_BRING_ME_HOME);
-
 				Intent addIntent = new Intent();
-				addIntent.putExtra(Intent.EXTRA_SHORTCUT_INTENT, shortcutIntent);
+				addIntent.putExtra(Intent.EXTRA_SHORTCUT_INTENT, MainActivity.getShortcutIntent(getContext()));
```
grote: Thanks for factoring out this code into its own method!

Just a bit strange that the SettingsFragment needs to call a method from the MainActivity. Maybe this should be move to TransportrUtils instead?

src/de/grobox/liberario/activities/MainActivity.java
```diff
@@ -91,6 +98,7 @@
 	private AccountHeader accountHeader;
 	private Toolbar toolbar;
 
+	@TargetApi(Build.VERSION_CODES.N_MR1)
```
grote: Why did you add this annotation? This method should work on other API levels as well, right?

If you want, you could put the shortcut creation code into its own method and annotate it like this.

res/values/strings.xml
patkan: I did not think of this and it really looks a bit weird. This text would be shown in launchers with lots of space. But if I think of it "Get Me Home" is more or less all that you need to know. I'll just completely remove that.

src/de/grobox/liberario/activities/MainActivity.java
patkan: I understand the annotation as a sign to the linter, that we aware of using API that is not supported by the minimumSDKVersion. An example is `ShotcutManager`. 

I will move that out because the rest of the `onCreate` should not be affected by that.

patkan:
I have only tested this in an API 25 emulator. There I did not see any problems, but since I have no phone with 7.1 I can't be 100% sure what to expect. (Nice feature: you can "pull" this shortcut out of this shortcut drawer onto the launcher, creating a nice-looking shortcut)
Maybe a 7.1-user could test this if needed? Maybe @maninthemiddle01 or @TheLastProject?

patkan:
Oh wait, I have found a bug :-) This menu is only available after the first run (because the onCreate of the MainActivity is not executed on installation). I will fix this and and then tell you.

grote:
> Oh wait, I have found a bug :-) This menu is only available after the first run (because the onCreate of the MainActivity is not executed on installation).

I am not sure this is a bug, because the user has to choose a transport network on first run. Without that, the home shortcut doesn't make sense and crash the app.

grote:
Is this good to go or do you want to change something?

patkan:
Sorry that I did not answer yet.
After having checked some alternative options I agree with you, it does not make much sense to have this before setting Transportr up for the first time. So I would say this is OK.

grote:
Thanks! :)

PR #311, opened 2017-02-07
[UNTESTED] Pluralize message

M1dgard:
I don't do much Android development yet and I haven't built the project, but I think this is more or less what has to change to allow plurals, as suggested by TheLastProject on Transifex a year ago.


grote:
Thanks!

PR #317, opened 2017-02-23
update pte

patkan:
* various fixes in AbstractProviders
* changes in API_BASE for VGN, DING, HSL, TLEM
* remove the dead provider in Transportr
* remove the "secret" for VGN as this is not needed anymore due to the PTE-update


grote:
Thanks a lot! :)

Any of these removed providers have drawables that need to be removed as well?

patkan:
Thanks, forgot to check the drawables :-) I will update my PR.

grote:
Thanks, the source file in art/networks is probably not needed anymore as well.

patkan:
Well, it's still in the git if we need it.
PACA had no drawable, neither could I find ATC, so this should be all of them.

PR #318, opened 2017-02-23
update pte (v1)

patkan:
* various fixes in AbstractProviders
* changes in API_BASE for VGN, DING, HSL, TLEM
* remove the dead provider in Transportr
* remove the "secret" for VGN as this is not needed anymore due to the PTE-update

This is cherry-picked from #317.


grote:
Thanks!

PR #319, opened 2017-02-23
fix NearbyStationsFragment when using GPS

### LINKED ISSUE #314: Searching for nearby stations by GPS does not show the results
Version: 1.1.7

When going to ‘Nearby Stations’ from the side menu and selecting ‘Use Current Position’, GPS is activated, and the location box is updated with the current GPS position. Network requests are then made, but the results are not shown on the screen.

This appears to be due to the following line in `NearbyStationsFragment`, since `deactivateGPS` is called from `LocationGpsView` when the GPS lookup completes:

```java
LocationGpsView.LocationGpsListener listener = new LocationGpsView.LocationGpsListener() {
	...
	@Override
	public void deactivateGPS() { ui.stations_area.setVisibility(GONE); }
```

Removing this statement fixes the issue and results are shown as expected.

patkan:
fixes #314

This uses the fix by @RunasSudo to show the results of a GPS-based
search in NearbyStationsFragment.

I messed a bit around with changing orientation, restarting the app, changing fragments and could not see any degradation.

### DIFF
--- a/src/de/grobox/liberario/fragments/NearbyStationsFragment.java
+++ b/src/de/grobox/liberario/fragments/NearbyStationsFragment.java
@@ -107,9 +107,7 @@ public class NearbyStationsFragment extends TransportrFragment {
 			}
 
 			@Override
-			public void deactivateGPS() {
-				ui.stations_area.setVisibility(GONE);
-			}
+			public void deactivateGPS() { }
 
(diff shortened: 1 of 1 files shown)


grote:
Wow, third PR in a day! :)

Thanks a lot!

Issue #329, opened 2017-03-14
Add favorites as home screen launcher (or as widget or shortcut)


moejoeblue:
Feature request: It would be great to have the possibility to save relations specified as favorites as launchers on the android home screen. (This feature has been removed from the DB Navigator app. It would be great to see it in TransportR!)
This could be done as a shortcut, similar to the "Get me Home"-shortcut that is already implemented (which is great).
The idea of a shortcut would be to save as much screen taps as possible: Ideally only one tap on the shortcut "Aa to Bb" should open TransportR and directly give the search results for Aa to Bb for the current time.

grote:
This has been implemented in #375 and is available in the latest beta. Thanks @Noxsense

PR #375, opened 2017-10-24
WIP: Add shortcut for normal favorite location

Noxsense:
Hello, I'm currently working on #329.
I tried to put the your new `addShortcut()` into the parent class, so the `FavoriteTripPopupMenu` can also create shortcuts, if they are marked as favorites.
I used the logic of `findDirection(...)` for the Shortcut Intent and checked if all parameters were correctly inserted, but the shortcut is only created (or added) if the parameters for `from`, `via` and `to` are null (uid and boolean were fine).

(The Intent created for the shortcut is actually working: I tried to start its activity instead of putting it as shortcut into the other intent.)

I have no idea the shortcut isn't showing up or only if the  parameters were null.
Do you have any idea why this happens?


app/src/main/java/de/grobox/transportr/utils/TransportrUtils.java
```diff
@@ -649,6 +649,22 @@ static public int getDragDistance(Context context) {
...
+		shortcutIntent.setAction(ACTION_MAIN);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_NEW_TASK);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_CLEAR_TOP);
+
+		//FIXME Doesn't work if they are not null
+		shortcutIntent.putExtra(FAV_TRIP_UID, uid);
+		shortcutIntent.putExtra(FROM, from);
+		shortcutIntent.putExtra(VIA, via);
+		shortcutIntent.putExtra(TO, to);
+		shortcutIntent.putExtra(SEARCH, false);
```
grote: My guess is that you can't put complex objects that need to be serialized into shortcut intents.

grote: If that's true, you could work with `StoredSearch` `uid`s, but you need to handle the case when the `uid` doesn't exist anymore.

Noxsense:
Hey, I've tried to store the locations ID ((instead of only StoredSearches)) and I could restore those from the intent.
But I think it' because the shortcut is trying to start searching immediately the database is not always loaded at time an there are no locations to restore from yet.
I used `getLocations` from the `DirectionViewModel` in the `DirectionActivity` but also the database itself isn't always loaded yet.

It's working if the shortcut is launched not long after quitting the app, I guess it's because of the cache.

Then I tried to load the already existing Home shortcut with cleared cache (or just updated app) and there the locations doesn't load either.

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -50,12 +50,14 @@ public boolean onMenuItemClick(MenuItem item) {
...
-				widgetName = String.format(widgetName, String.valueOf(trip.getFrom()), String.valueOf(trip.getTo()));
-
-				addShortcut(context, getShortcutIntent(context, 0, trip.getFrom(), trip.getVia(), trip.getTo()),
-						widgetName, R.mipmap.ic_quickwork_widget);
+				addShortcut(context, getShortcutIntent(context,
+							trip.getUid(),
+							trip.getNetworkId(),
+							((StoredLocation) trip.getFrom()).getUid(),
+							(trip.getVia() == null) ? null : ((StoredLocation) trip.getVia()).getUid(),
+							((StoredLocation) trip.getTo()).getUid()),
```
grote: `trip.getUid()` should be sufficient to identify a trip and get the `FavoriteTripItem` from the database. No need for the other data here.

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -50,12 +50,14 @@ public boolean onMenuItemClick(MenuItem item) {
...
+						getShortcutName(),
+						R.mipmap.ic_quickwork_widget);
```
grote: When the PR is ready, we would need to think about  a different icon here.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -126,6 +128,30 @@ private void processIntent() {
...
+					fromId = intent.getLongExtra(FROM, 0);
+					viaId = (Long) intent.getSerializableExtra(VIA);
+					toId = intent.getLongExtra(TO, 0);
+
+					from = null;
+					to = null;
+
+					try{
+						for (FavoriteLocation location : viewModel.getLocations().getValue()) {
+							locationId = location.getUid();
```
grote: You need to ask the view model for the FavoriteTripItem with the `uid` from the intent. If the data is not yet available when the intent is processed, you need to observe it to get notified when available.

app/src/main/java/de/grobox/transportr/utils/TransportrUtils.java
```diff
@@ -649,19 +650,18 @@ static public int getDragDistance(Context context) {
...
 		shortcutIntent.putExtra(FAV_TRIP_UID, uid);
-		shortcutIntent.putExtra(FROM, from);
-		shortcutIntent.putExtra(VIA, via);
-		shortcutIntent.putExtra(TO, to);
-		shortcutIntent.putExtra(SEARCH, false);
-
+		shortcutIntent.putExtra("network", networkId.ordinal());
+		shortcutIntent.putExtra(FROM, fromId);
+		shortcutIntent.putExtra(VIA, viaId);
+		shortcutIntent.putExtra(TO, toId);
```
grote: like I said above, the `FAV_TRIP_UID` should be sufficient here.

Noxsense:
The name for the shortcut is currently hard coded. But if all names from to and via have to be inserted, it will be quite long.
And I think the icon could the star like the favourite icon  on a red background.

grote:
> The name for the shortcut is currently hard coded. But if all names from to and via have to be inserted, it will be quite long.

Yeah, I don't know how to solve that.

> And I think the icon could the star like the favourite icon on a red background.

I agree!

comradekingu:
@Noxsense Could you post a screenshot of what that looks like?

Noxsense:
I guess something like this?
ic_favourite_generated


[Icon (with Launcher Icon generator)](https://romannurik.github.io/AndroidAssetStudio/icons-launcher.html#foreground.type=clipart&foreground.clipart=star&foreground.space.trim=1&foreground.space.pad=0.15&foreColor=rgb(255%2C%20255%2C%20255)&backColor=rgb(242%2C%2066%2C%2054)&crop=0&backgroundShape=square&effects=shadow&name=ic_favorite_widget)

grote:
Yes looks nice. Android Studio has a built-in icon creator now btw.

Noxsense:
Is there anything else I can improve?

app/src/main/java/de/grobox/transportr/favorites/trips/AbstractFavoritesPopupMenu.java
```diff
@@ -64,4 +71,22 @@ public boolean onMenuItemClick(MenuItem item) {
 		}
 	}
 
+	protected void addShortcut(Context context, Intent shortcutIntent, String shortcutName, int shortcutDrawable) {
```
grote: Do you really need to pass in a `Context` here or can you use the existing `context` variable?

The `shortcutDrawable` should be annotated with `@DrawableRes`.

app/src/main/java/de/grobox/transportr/favorites/trips/AbstractFavoritesPopupMenu.java
```diff
@@ -64,4 +71,22 @@ public boolean onMenuItemClick(MenuItem item) {
...
+		context.sendBroadcast(addIntent);
+
+		// switch to home-screen to let the user see the new shortcut
+		Intent startMain = new Intent(Intent.ACTION_MAIN);
+		startMain.addCategory(Intent.CATEGORY_HOME);
+		startMain.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+		context.startActivity(startMain);
+	}
+
+	protected abstract Intent getShortcutIntent(Context context, String type);
```
grote: Why is this needed if you pass in the shortcut intent above?

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -20,17 +20,27 @@
...
 import android.view.View;
 
 import de.grobox.transportr.R;
+import de.grobox.transportr.trips.search.DirectionsActivity;
+
+import static android.content.Intent.ACTION_SEARCH;
+import static android.content.Intent.FLAG_ACTIVITY_NEW_TASK;
+import static android.content.Intent.FLAG_ACTIVITY_CLEAR_TOP;
+
+//import static de.grobox.transportr.utils.Constants.FAV_TRIP_UID;
```
grote: Is this still needed?

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -67,4 +80,32 @@ private void setFavState(MenuItem item, boolean is_fav) {
...
+	@Override
+	protected Intent getShortcutIntent(Context context, String type)
+	{
+		// ignore type, insert uid
+		Intent shortcutIntent = new Intent(context, DirectionsActivity.class);
+		shortcutIntent.setAction(ACTION_SEARCH);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_NEW_TASK);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_CLEAR_TOP);
+		shortcutIntent.setData(Uri.parse(DirectionsActivity.INTENT_URI_FAVORITE));
+		shortcutIntent.putExtra("uid", trip.getUid());
```
grote: `"uid"` should be a `String` constant.

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -67,4 +80,32 @@ private void setFavState(MenuItem item, boolean is_fav) {
...
+		return shortcutIntent;
+	}
+
+	private String getShortcutName()
+	{
```
grote: coding style violation (here and below in this method)

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -67,4 +80,32 @@ private void setFavState(MenuItem item, boolean is_fav) {
...
+		String str = null;
+
+		if (trip.getVia()==null)
```
grote: missing spaces

app/src/main/java/de/grobox/transportr/favorites/trips/SavedSearchesViewModel.java
```diff
@@ -42,7 +42,7 @@ protected SavedSearchesViewModel(TransportrApplication application, TransportNet
 		this.savedSearches = searchesRepository.getFavoriteTrips();
 	}
 
-	LiveData<List<FavoriteTripItem>> getFavoriteTrips() {
+	public LiveData<List<FavoriteTripItem>> getFavoriteTrips() {
```
grote: Why did you need to change the visibility here? Do you access this ViewModel outside of this package?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
 					to = viewModel.getWork().getValue();
 					if (to == null) new WorkPickerFragment().show(getSupportFragmentManager(), WorkPickerFragment.TAG);
 					break;
+				case INTENT_URI_FAVORITE:
+					from = null;
+					via = null;
+					to = null;
```
grote: Do you need to explicitly assign `null` here?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
+					long uid = intent.getLongExtra("uid",0);
```
grote: Missing space. Btw. in Android Studio, there's a shortcut to automatically apply the current coding style to the entire file or your selection. This quickly fixes the style.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
+
+					LiveData<List<FavoriteTripItem>> tripsLiveData;
+					tripsLiveData = viewModel.getFavoriteTrips();
+					tripsLiveData.observe(this, trips -> {
+						for (FavoriteTripItem trip : trips) {
+							if (trip!=null && trip.getUid()==uid)
+							{
```
grote: coding style

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
 					break;
+				case INTENT_URI_FAVORITE:
+					from = null;
+					via = null;
+					to = null;
+					long uid = intent.getLongExtra("uid",0);
+
+					LiveData<List<FavoriteTripItem>> tripsLiveData;
+					tripsLiveData = viewModel.getFavoriteTrips();
+					tripsLiveData.observe(this, trips -> {
```
grote: You could just do `viewModel.getFavoriteTrips().observe(...)`

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
+							if (trip!=null && trip.getUid()==uid)
+							{
+								searchFromTo(trip.getFrom(), trip.getVia(), trip.getTo());
+								tripsLiveData.removeObservers(this); // everything is done
+								break;
+							}
+						}
+					});
+					setIntent(null);
+					return; // will be finished in viewmodel's loaded trips
```
grote: This looks a bit strange. Why don't you let this method go down to the end and assign `from`, `via` and `to` properly instead of suddenly returning here and duplicating code before?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
+					tripsLiveData = viewModel.getFavoriteTrips();
+					tripsLiveData.observe(this, trips -> {
+						for (FavoriteTripItem trip : trips) {
+							if (trip!=null && trip.getUid()==uid)
+							{
+								searchFromTo(trip.getFrom(), trip.getVia(), trip.getTo());
+								tripsLiveData.removeObservers(this); // everything is done
+								break;
+							}
+						}
```
grote: What happens if the user deleted a favorite trip, but still has the shortcut on the launcher? How is this handled?

app/src/main/java/de/grobox/transportr/favorites/trips/SavedSearchesViewModel.java
Noxsense: Yes, in `DirectionsActivity`
where I use it like `viewModel.getFavoriteTrips().observe(...)`

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
Noxsense: Yes, but I use `tripsLiveData` to remove the observer when the actual trip is found.
Otherwise the found trip will be found again on every update of the livedata with the same trips.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
Noxsense: The shortcut will just start a `DirectionsActivity` with empty source and destination fields. The observer won't find a matching trip in the list and do nothing.
Should we try to delete shortcuts when unmarking trips as favorite?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
mheinzel: The main problem is that we don't know when all trips will be loaded and if they even contain the one we are looking for. So if we just use `getValue()` (as for Home or Work) it might not be loaded in time and we miss it. But we can't block here and wait for it either, since it might never appear.
The only solution I saw, was to attach `searchFromTo` to the observer and exit the method. But I don't have a valid `from` and `to` at this point, so I can't let the method go down to the end.

If you have a better idea, I would be happy to hear it.

I'll fix the other issues you pointed out.

mheinzel: Oh, and thank you for taking the time to review this!

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
grote: Ah right. You need a reference to the LiveData.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
grote: I think when the observer doesn't find any trips, we could just show a Toast saying that this favorite was removed and that the user should remove the shortcut as well.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
```diff
@@ -147,6 +152,26 @@ private void processIntent() {
...
+					from = null;
+					via = null;
+					to = null;
+					long uid = intent.getLongExtra("uid",0);
+
+					LiveData<List<FavoriteTripItem>> tripsLiveData;
+					tripsLiveData = viewModel.getFavoriteTrips();
+					tripsLiveData.observe(this, trips -> {
+						for (FavoriteTripItem trip : trips) {
+							if (trip!=null && trip.getUid()==uid)
```
grote: `trip` should never be `null` here.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsActivity.java
Noxsense: The main problem is that we don't know when all trips will be loaded and
if they even contain the one we are looking for. So if we just use
`getValue()` (as for Home or Work) it might not be loaded in time and we
miss it. But we can't block here and wait for it either, since it might
never appear.
The only solution I saw, was to attach `searchFromTo` to the observer
and exit the method. But I don't have a valid `from` and `to` at this
point, so I can't let the method go down to the end.

If you have a better idea, I would be happy to hear it.

Thank you for taking the time to review my changes.
I'll fix the other
issues you pointed out.

grote: Ah I see. Unfortunately, I don't see a better way to handle this myself.

app/src/main/java/de/grobox/transportr/favorites/trips/AbstractFavoritesPopupMenu.java
Noxsense: Previously, it was necessary to call `getShortcutIntent()` to get the intent and pass it to `addShortcut()`.
This is unnecessary, so I changed `addShortcut()` to call `getShortcut()` itself.

`getShortcutIntent()` is needed because each subclass has its own way of creating the actual shortcut intent.

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -47,6 +57,9 @@ public boolean onMenuItemClick(MenuItem item) {
 				}
 				setFavState(item, trip.isFavorite());
 				return true;
+			case R.id.action_add_shortcut:
+				addShortcut(null, getShortcutName(), R.mipmap.ic_launcher_favorite);
```
grote: What is a `null` type? If `null` makes sense here, you need to annotate the parameter `@Nullable`.

app/src/main/java/de/grobox/transportr/favorites/trips/FavoriteTripPopupMenu.java
```diff
@@ -67,4 +80,21 @@ private void setFavState(MenuItem item, boolean is_fav) {
...
+	protected Intent getShortcutIntent(Context context, String type) {
+		// ignore type, insert uid
+		Intent shortcutIntent = new Intent(context, DirectionsActivity.class);
+		shortcutIntent.setAction(ACTION_SEARCH);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_NEW_TASK);
+		shortcutIntent.addFlags(FLAG_ACTIVITY_CLEAR_TOP);
+		shortcutIntent.setData(Uri.parse(DirectionsActivity.INTENT_URI_FAVORITE));
+		shortcutIntent.putExtra(FAV_TRIP_UID, trip.getUid());
+		return shortcutIntent;
+	}
```
grote: Could you also pass in the proper type here and then do

```java
Intent shortcutIntent = super.getShortcutIntent(Context context, String type);
shortcutIntent.putExtra(FAV_TRIP_UID, trip.getUid());
return shortcutIntent;
```

This way, you can maybe get rid of the code duplication.

Noxsense: Ok, now I am reusing some more of the methods which where actually in `SpecialLocationPopupMenu` in `AbstractFavoritesPopupMenu`, while most of them are still abstract and are overridden in the actual sub class. So all of them use `addShortcut()` and `getShortcutIntent()` with `FavoriteTripPopupMenu` adding the trips `uid`.

grote:
Merged, thanks!

Noxsense:
Nice. Thanks for your help!

PR #377, opened 2017-11-05
Australia navitia

pserwylo:
Add Australia provider based on Navitia. Specified "Beta", given it doesn't include live data and I've only substantially tested Melbourne, Victoria. I have tried to test the rest of the localities, but I don't live in those areas and so can't comment on how actually useful they are.


grote:
Thanks @pserwylo! This is looking good!

PR #388, opened 2017-12-02
Use correct language for 9292 provider

### LINKED ISSUE #386: 2.0.0-beta1 search weirdness with 9292
The search in 2.0.0-beta1 seems a bit broken, at least with 9292:

As an example: "bushalte heuvel" does not return a single bus stop with this name, yet "heuvel bushalte" returns all of them. I can't seem to reproduce this with other bus stops.

I also noticed all results are now Dutch. While I previously searched for "bus stop" I now have to search for "bushalte", regardless of device language. This was unexpected to me.

full-duplex:
This fixes #386.


grote:
Thanks!

PR #389, opened 2017-12-05
Add support for Nicaragua providers - v2

xamanu:
Please add support and logos for providers in Nicaragua.


grote:
> this can now be submitted as a PR upstream.

Sorry, unfortunately it is not quite that easy as the commit contains unrelated changes as well and changes no longer necessary.

xamanu:
Sorry, my fault, I rebased.

grote:
The XML representation of the logo is huge (compared to the others). Couldn't the SVG be simplified a bit? Just the stripes and maybe a simple triangle in the middle? I would also suggest to remove all the groups from the SVG and join paths where possible, before doing the conversation to XML.

Furthermore, there's no Nicaragua provider in Transportr, yet. Maybe the logo should be submitted together with the provider?

You might also want to consider whether there should be one single Nicaragua Provider covering several cities or one provider per city. Changing this later is inconvenient for users as they need to re-select their provider and lose all saved data.

xamanu:
> The XML representation of the logo is huge (compared to the others). Couldn't the SVG be simplified a bit? Just the stripes and maybe a simple triangle in the middle? I would also suggest to remove all the groups from the SVG and join paths where possible, before doing the conversation to XML.

Yes, I simplified it.

> Furthermore, there's no Nicaragua provider in Transportr, yet. Maybe the logo should be submitted together with the provider?

It's actually in the pipeline: https://github.com/schildbach/public-transport-enabler/pull/181

> You might also want to consider whether there should be one single Nicaragua Provider covering several cities or one provider per city. Changing this later is inconvenient for users as they need to re-select their provider and lose all saved data.

Currently we don't have enough coverage and some unresolved problems to provide a Nicaragua-wide GTFS.

xamanu:
I know that @altnico and @ialokim also suggested to have one for whole Nicaragua. I guess it is worth to talk together about it (here?):

Advantages of a country wide GTFS:
* Easier for the user to select
* Extensions and improvements to the GTFS are going to be included almost on the fly.

Disadvantages of a country wide GTFS:
* More difficult to maintain.
* On the current stage there is only the capital, and one middle size city and the bus connecting it. We have an unresolved problem on places (stops) with the same name in different cities.

I think it is not realistic to think, that if we go the path for one for the whole Nicaragua it will be soon anywhere close to complete. However it would probably help to work on it. I'm really not sure about which direction to take. What do you think?

xamanu:
I extended this pull request with the other parts that seem to be necessary in Transportr to support Managua.

app/src/main/res/drawable/network_nimanagua_logo.xml
```diff
@@ -0,0 +1,33 @@
...
+        android:fillColor="#ffffff"
+        android:strokeWidth="0.082623"
+        android:pathData="M 0 21.333668 H 64 V 42.666668 H 0 V 21.333668 Z" />
+    <path
+        android:fillColor="#c8a400"
+        android:strokeWidth="0.76141131"
+        android:pathData="M 32.835766 22.1992101 C 38.1324754817 22.1992101 42.4263069 26.4930415183 42.4263069 31.789751 C 42.4263069 37.0864604817 38.1324754817 41.3802919 32.835766 41.3802919 C 27.5390565183 41.3802919 23.2452251 37.0864604817 23.2452251 31.789751 C 23.2452251 26.4930415183 27.5390565183 22.1992101 32.835766 22.1992101 Z" />
+    <path
+        android:fillColor="#ffffff"
+        android:strokeWidth="1.2304405"
```
grote: Way better smaller file! You could also remove these `android:strokeWidth` from the XML. They don't do anything on Android.

app/src/main/java/de/grobox/transportr/networks/Region.java
```diff
@@ -56,6 +56,7 @@
 	CANADA(R.string.np_region_canada, "🇨🇦"),
 	COSTA_RICA(R.string.np_region_costa_rica, "🇨🇷"),
 	AFRICA(R.string.np_region_africa, "🌍");
+	NICARAGUA(R.string.np_region_nicaragua, "🌍");
```
grote: Here, you should add the flag of Nicaragua as a second parameter.

app/src/main/res/xml/changelog_master.xml
```diff
@@ -10,6 +10,7 @@
 		<change>Addresses will be saved</change>
 		<change>Round icon for new Android versions</change>
 		<change>Added New York, Ghana, Costa Rica</change>
+		<change>Added Managua</change>
```
grote: Please add this to the list one line above, since this will only be shown to users when they upgrade from transportr 1 to 2.

app/src/main/java/de/grobox/transportr/networks/Region.java
ialokim: It's the UTF8-Representation of the flag, see https://emojipedia.org/flag-for-nicaragua/

xamanu:
* Use proper logo for Managua (based on rutas.mapanica.net).
* Added also Esteli to this pull request.

app/src/main/java/de/grobox/transportr/networks/TransportNetworks.java
```diff
@@ -635,6 +635,25 @@
...
+					.setId(NetworkId.MANAGUA)
+					.setName(R.string.np_name_managua)
+					.setDescription(R.string.np_desc_managua)
+					.setRegion(Region.NICARAGUA)
+					.setStatus(ALPHA)
+					.setGoodLineNames(true)
+					.build(),
+
+			new TransportNetworkBuilder()
+					.setId(NetworkId.ESTELI)
```
grote: This PR will probably fail CI as these `NetworkId`s are not yet included in the version of PTE included here.

app/src/main/res/drawable/network_niesteli_logo.xml
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="60.000006dp"
+    android:height="59.999998dp"
+    android:viewportWidth="60.000006"
+    android:viewportHeight="59.999998">
+
+    <group
+            android:translateX="-1.1721562"
+            android:translateY="-210.77028">
```
grote: Could you maybe remove the group here (from the SVG and then export again). Also please use exactly `64dp` for height and width.

app/src/main/res/drawable/network_niesteli_logo.xml
```diff
@@ -0,0 +1,107 @@
...
+        <path
+            android:fillColor="#ffffff"
+            android:fillAlpha="0.98999999"
+            android:strokeAlpha="0.98999999"
```
grote: These alphas can be removed. They don't do anything on Android.

app/src/main/java/de/grobox/transportr/networks/Region.java
```diff
@@ -56,6 +56,7 @@
 	CANADA(R.string.np_region_canada, "🇨🇦"),
 	COSTA_RICA(R.string.np_region_costa_rica, "🇨🇷"),
 	AFRICA(R.string.np_region_africa, "🌍");
```
ialokim: has to end with a comma here because the enum continues in the next line

app/src/main/java/de/grobox/transportr/networks/TransportNetworks.java
```diff
@@ -635,6 +635,25 @@
 					.setGoodLineNames(true)
 					.build(),
 
+			// Nicaragua
+
+			new TransportNetworkBuilder()
+					.setId(NetworkId.MANAGUA)
+					.setName(R.string.np_name_managua)
```
AltNico: Where do you defined this strings?

app/src/main/java/de/grobox/transportr/networks/NetworkProviderFactory.java
```diff
@@ -867,6 +869,26 @@ public static synchronized NetworkProvider provider(final NetworkId networkId) {
...
+				final ManaguaProvider provider = managuaProviderRef.get();
+				if (provider != null)
+					return provider;
+			}
+
+			final ManaguaProvider provider = new ManaguaProvider(NAVITIA);
+			managuaProviderRef = new SoftReference<>(provider);
+			return provider;
+		} else if (networkId.equals(NetworkId.ESTELI)) {
+			if (esteliaProviderRef != null) {
```
ialokim: typo, did you mean `esteliProviderRef`?

app/src/main/java/de/grobox/transportr/networks/Region.java
xamanu: Yes, indeed.

app/src/main/java/de/grobox/transportr/networks/NetworkProviderFactory.java
xamanu: Yes, thanks for noticing!

xamanu:
Thanks for your reviews, @grote and @ialokim! I fixed based on your comments.

app/src/main/java/de/grobox/transportr/networks/TransportNetworks.java
xamanu: In the strings.xml (now added after a comment by @ialokim)

AltNico:
So I was able to build the app by changing the pte dependency to [nicaragua-providers--merged-into-otm](https://github.com/mapanica/public-transport-enabler/tree/nicaragua-providers--merged-into-otm). However, it does not find any information in Managua and Estelí. I think this is caused by pte looking for information at api.navitia.io instead of navitia.mapanica.net.

grote:
Yes, you should make sure your providers don't use the default API URL.

Before this can be merged, the upstream PTE MR needs to be ready (tests still missing) and then be merged into the staging PTE.

xamanu:
@grote, can you please point me to the tests which are missing? I guess there are other providers that implement such.

grote:
The live tests in your PTE pull requests are still missing. If it is in the PTE staging repo already anyway, it could already be integrated into transportr.

xamanu:
I just included it into the staging PTE. This PR needs to be modified to fit the single Nicaragua layer, as it was included in navitia.io

xamanu:
Closing in favour of its improved and working follower: #419

PR #390, opened 2017-12-07
Please rename region 'Africa' to 'Ghana' to keep consistency

ialokim:
As [commented on the commit](https://github.com/grote/Transportr/commit/a5123191f8b5a323b708949a87b275483d1ac1c3#r26095681), I think it's much better to always use country names as regions (if it doesn't contain cross-border networks) instead of continent names.

As in the cases of Belgium, Italia and Spain, it would be a region called `Ghana` with one network called `Ghana`, too.

### DIFF
--- a/app/src/main/java/de/grobox/transportr/networks/Region.java
+++ b/app/src/main/java/de/grobox/transportr/networks/Region.java
@@ -55,7 +55,7 @@ enum Region {
 	BRAZIL(R.string.np_region_br, "🇧🇷"),
 	CANADA(R.string.np_region_canada, "🇨🇦"),
 	COSTA_RICA(R.string.np_region_costa_rica, "🇨🇷"),
-	AFRICA(R.string.np_region_africa, "🌍");
+	GHANA(R.string.np_region_ghana, "🇬🇭");
 
 	private final @StringRes int name;
 	private final @Nullable String flag;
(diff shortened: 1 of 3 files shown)


grote:
Like explained in https://github.com/grote/Transportr/commit/a5123191f8b5a323b708949a87b275483d1ac1c3#r26095779 I do not intent to change this. The main reason for grouping providers into expandable regions is to not have a super long list where it is hard to find stuff, especially with the many providers in Germany when people are interested in a different country.

Over time this became inconsistent, but the goal is to avoid regions with just one provider. For Costa Rica, we could think about introducing a Latin America Region and put Brazil in there as well. If multiple nesting levels are easily possible, we could also do a continent -> country -> provider sorting.

ialokim:
> If multiple nesting levels are easily possible, we could also do a continent -> country -> provider sorting.

I would go for this if it's possible. I'll look into it in the next few days if you're okay.

But I definitely disagree with the current state, because mentioning Africa being the only continent on the same level as (mostly Northern) countries could be malinterpreted considering human's history.

grote:
> I'll look into it in the next few days if you're okay.

That'd be ok with me.

ialokim:
> That'd be ok with me.

Here's my pull request #395.

PR #296, opened 2017-01-22
Add support for California transit systems.

n76:
Finally figured out how to build this project. :)

Along with changes in the [public-transport-enabler](https://github.com/schildbach/public-transport-enabler/pull/133), this seems to work at least as far as my local Orange County Transit Authority (OCTA) system is concerned. I haven't yet checked for the other transit systems in the source data.

Signed-off-by: Tod Fitch <Tod@FitchDesign.com>

PR #392, opened 2017-12-09
Add California transit systems

n76:
Signed-off-by: Tod Fitch <Tod@FitchDesign.com>

Replacement for earlier pull request. This one should apply cleanly to current master branch.


n76:
https://github.com/grote/Transportr/pull/296 is already closed. Sorry about the translations, should I revise the pull request to remove them?

grote:
Yes, I think that would be better. Why add something that gets removed anyway.

grote:
@n76 the translators have a question for you: https://www.transifex.com/grobox/transportr/translate/#fr/strings/128628594

n76:
@grote That link leads to a 404 page not found error for me. I signed up for an account at Transiflex and looked that the Transportr project. I don't see how to look at comments and it shows French as 100%. When I tried to join the project team it required me to select from a limited list of languages that did not include any that I know.

grote:
Argh. I am sorry about this. Transifex really sucks. Does it work if you replace `#fr` with `#en` in the URL?

Yahoe001:
@n76 @grote We have a typo in "Global Transport in Calfornia State (USA)". It should read California instead. Also, is "Global Transport " meant here as a company name or in a general sense?

Yahoe001:
@grote @n76 Only project managers/maintainers and specific language team members can use the Transifex comment/issue system. So no Transifex doesn’t "suck" as you put it, it is the way it built.

n76:
@grote @Yahoe001 Yes, correct the spelling of California in the English (or I can create another pull request to the Transport GitHub repository for it).

The "Global Transport" phrase was copied from the equivalent for the state of Oregon. There are multiple transit agencies in California. The change to Transportr allows the use of and routing over multiple agencies (Metrolink Rail, LA Metro, Orange County Transit Authority, CalTrain, etc.).

Looking at how other countries and areas in the US are done, it seems that when I copied the setup for Oregon I copied an outlier. I think there are two options here: Get the current list of providers in California from Navita and list them instead of the string "global transport". Or, use a phrase like "Most of California" or "Public transport in California".

@grote Looking at https://navitia.opendatasoft.com/explore/dataset/us-ca/table/?sort=type_file I am not sure that all the agencies shown are also in the "Global Transport in California State" file. Do you know of an easy way to verify that or not? If they aren't, is there a way through the PTE to combine them for use in Transportr?

grote:
> I am not sure that all the agencies shown are also in the "Global Transport in California State" file.

They should be. However, some have a validity date in the past. These won't work until updated.

> Do you know of an easy way to verify that or not?

You can only try to search for departures or connections covered by those agencies' GTFS.

> If they aren't, is there a way through the PTE to combine them for use in Transportr?

The API to access them is provided by navitia.io. PTE is just accessing this API. If you don't want to run your own Navitia server, there's not much we can change here.

The best course of action is to see if updated GTFS files for the outdated files are available and if so, ping navitia to update them. If no updates are available, the respective agencies could be pinged.

n76:
Created new pull request to improve text. https://github.com/grote/Transportr/pull/396

PR #395, opened 2017-12-12
Order regions by continents

ialokim:
This PR adds the ordering of transport network regions into continents and therefore an easier way of picking the requested region as suggested [here](https://github.com/grote/Transportr/pull/390#issuecomment-350249478) by @grote.

It is already working on my personal phone, here some screenshots:
screenshots

Before merging it in, I think we should discuss the following points:
* Should the continents present little icons at the left, something like [this](http://www.endlessicons.com/free-icons/africa-icon/) (see also their [Terms of Use](http://www.endlessicons.com/terms-of-use/)? Should they be in SVG or PNG format?
* As almost all `Region`s represent countries, shouldn't we rename them in code to `Country`?
* The continent list item design is, as for now, almost identical to (because copied of) the region's design. Any ideas on how to improve the distinction are very welcome!
* Should we use the [Wikipedia definition](https://en.wikipedia.org/wiki/Central_America) where Central America is part of North America or should we separate it as another subcontinent of America as defined in [Ibero-America](https://en.wikipedia.org/wiki/Central_America#cite_note-6)?
* I think it would be necessary to translate the continent names in the respective `strings.xml`


grote:
Thanks a lot for working on this!

> Should the continents present little icons at the left, something like this (see also their Terms of Use? 

I think little recognizable icons would be nice. They should be under a free license.
You should be able to re-use the `RegionViewHolder` if you add icons, right?

> Should they be in SVG or PNG format?

In Android XML vector drawable format. SVG can usually be converted.

> As almost all Regions represent countries, shouldn't we rename them in code to Country?

Which ones do not represent countries?

I personally think, we only need a region/country if there's more than a few providers in it. Otherwise, we can add the providers directly at the continent level, especially when they are supposed to cover an entire country anyway.

> Should we use the Wikipedia definition where Central America is part of North America or should we separate it as another subcontinent of America as defined in Ibero-America?

To me it feels odd that Costa Rica is part of North America. Maybe use [Middle America](https://en.wikipedia.org/wiki/Middle_America_(Americas)) instead?

ialokim:
> I think little recognizable icons would be nice. They should be under a free license.

If I understand it right, in the proposed case there's no official licence, but 'free to use (...) for use in both personal and commercial projects' as stated [here](http://www.endlessicons.com/terms-of-use/). Btw, it features an 'Central American' icon which includes Mexico...

> You should be able to re-use the RegionViewHolder if you add icons, right?

Yes, but maybe we should rename it to avoid confusing?

> Which ones do not represent countries?

Only `Europe` (with cross-border trains) and `Africa` (known concern) for now.

> I personally think, we only need a region/country if there's more than a few providers in it. Otherwise, we can add the providers directly at the continent level, especially when they are supposed to cover an entire country anyway.

I understand your point, but that would be a little more complicated because the FastAdapter library with their expandable extention seems to support only one type of children (in this case would be _or_ `RegionItem` _or_ `TransportNetworkItem`). We should have an abstract parent class, not sure if this would work. But I'll look into it.

> To me it feels odd that Costa Rica is part of North America. Maybe use Middle America instead?

Personally, I don't find it very intuitive because I would never search Colombia or Venezuela in Middle America as stated on Wikipedia. Why not keeping Central America for Mexiko down to Panama? (Also see icon response).

ialokim:
> Did you check https://materialdesignicons.com/ and https://emojipedia.org/ for alternative icons?

I've checked them just now, they only have globes with [Europe and Africa](https://emojipedia.org/earth-globe-europe-africa/), [Asia and Australia](https://emojipedia.org/earth-globe-asia-australia/) or [North and South America](https://emojipedia.org/earth-globe-americas/) together.

What about Creative Commons Attribution-Share Alike 4.0 International, 3.0 Unported, 2.5 Generic, 2.0 Generic and 1.0 Generic license as in [this file](https://commons.wikimedia.org/wiki/File:AfricaContour.svg) of wikipedia. Seems like there could be an SVG for each continent we would like to include.

> OK, but these are continents now. We don't need those as Regions/Countries anymore.

That's right, if it's possible to have two different child element types.

grote:
CC-SA is fine. The contour lines probably would need to be simplified a bit before converted into a vector drawable, but this should work.

ialokim:
Working on it, I found a [bug in the FastAdapter library](/mikepenz/FastAdapter/pull/588), waiting for the merge to continue working on this feature.

ialokim:
I've already proceeded to include the continents contour lines (I've generated them myself from [this Wikimedia file](https://commons.wikimedia.org/wiki/File:BlankMap-World6.svg), published as "gemeinfrei") and changed my commit as requested to allow national transport networks to be shown as child elements of continents.

I've too worked a bit on the list (item) design to get a more ordered look and feel. See these screenshots of the current version:
out

As mentioned, there's still the error because of the bug in the FastAdapter library, but I thought of pushing it already to be able to discuss the changes.

ialokim:
The bugfix [is now merged](/mikepenz/FastAdapter/pull/588), I've updated the libraries.

grote:
Awesome! I'll try to find time for a review and test soon, but no promises.

ialokim:
> Awesome! I'll try to find time for a review and test soon, but no promises.

That would be great, but no need to hurry. Make sure you'll have a great talk at 34C3! Greetings by Nico, too.

app/build.gradle
```diff
@@ -113,8 +113,8 @@ dependencies {
 	implementation 'com.google.android:flexbox:0.3.1'
 	implementation 'com.mikepenz:materialdrawer:6.0.0'
 	implementation 'com.mikepenz:aboutlibraries:6.0.0'
-	implementation 'com.mikepenz:fastadapter-commons:3.0.3@aar'
-	implementation 'com.mikepenz:fastadapter-extensions-expandable:3.0.3@aar'
```
grote: Would you mind upgrading this to the latest `3.2.0`? The CI fails for `3.0.5` for some reason.

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
```diff
@@ -74,13 +74,13 @@ class PickTransportNetworkActivityTest : ScreengrabTest() {
         makeScreenshot("1_FirstStart")
 
         // hack to find region position in list
-        val regionList = ArrayList(EnumSet.allOf(Region::class.java))
+        val regionList = ArrayList(EnumSet.allOf(Country::class.java)) //TODO: pretty sure it will break :/
```
grote: If you are pretty sure this will break, is this PR then even ready for review or did you just forget to remove this TODO?

app/src/main/java/de/grobox/transportr/networks/TransportNetworkItem.java
```diff
@@ -56,9 +62,9 @@ public int getLayoutRes() {
 	}
 
 	@Override
-	public void bindView(TransportNetworkViewHolder ui, List<Object> payloads) {
+	public void bindView(RegionViewHolder ui, List<Object> payloads) {
```
grote: This looks fishy. Why would this be called with a `RegionViewHolder` when there's a `TransportNetworkViewHolder` as well?

app/src/main/res/drawable/continent_africa.xml
```diff
@@ -0,0 +1,14 @@
...
...0zm3.8194 16.99088c-3.5523 2.63503 -0.065 3.50123 0 0zm-68.6599 2.64746c4.9419 4.37319 -1.9493 3.98343 0 0zm24.3856 13.10142c12.5353 9.34666 0.5533 45.43428 2.8507 14.60687 0.7082 -4.97042 0.5552 -10.54564 -2.8507 -14.60687zm44.8752 10.70237l-0.482 0.30572 0.482 -0.30572zm-3.9745 2.43303c-2.2037 1.31735 -1.6964 1.01412 0 0zm32.7228 1.63052c-22.6121 13.30958 22.0257 16.76755 0 0zm-43.508 0.19744l0.2221 0.0902 -0.2221 -0.0902zm3.8003 0.95325l-0.1 0.0446 0.1 -0.0446zm-5.711 0.57111l-0.2335 0.29935 0.2335 -0.29935zm-15.4962 0.27387l-0.6878 0.35668 0.6878 -0.35668zm-4.6347 1.13797l0.8917 0.0297 -0.8917 -0.0297zm14.4199 0.41399l0.3396 0.30147 -0.3396 -0.30147zm3.4541 0.34769c-2.8605 0.69679 -1.7203 0.41906 0 0zm-9.5728 0.20382l1.484 0.16984 -1.484 -0.16984zm98.0325 2.32688c-5.4679 22.61953 -24.3092 31.25113 -40.2439 41.82703 8.4251 25.76743 -24.6727 57.47949 4.3408 75.82891 28.3554 -7.87684 23.0665 -45.38599 35.0182 -66.98526 4.9318 -19.09082 13.7802 -31.7672 0.8849 -50.67068zm-7.7131 10.56648c-2.1864 1.73758 3.4233 2.55496 0 0zm11.8106 30.17086c-2.2079 1.70654 -0.4682 4.86705 0 0zm-103.6331 18.31355c-1.558 -0.53209 0.8265 1.96713 0 0zm-7.297 23.61696c-0.4843 3.0056 -0.3419 2.12177 0 0z"
```
grote: Could you please simplify this massive path, remove the translation group around it and use the proper heigth/width values already here? This is best done in Inkscape before creating the XML.

grote: Please do the same for the other new drawables.

app/src/main/res/layout/list_item_transport_network.xml
```diff
@@ -11,8 +11,8 @@
 
 	<ImageView
 		android:id="@+id/logo"
-		android:layout_width="wrap_content"
-		android:layout_height="wrap_content"
+		android:layout_width="64dp"
```
grote: This won't look good for PNG images that we still have a few of. Better leave it as it was before.

app/build.gradle
ialokim: Yeah, no problem. 😄  It's already at `3.2.2` 🙊

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
ialokim: I'll look into Android tests, I've never worked with them before. But if you could give me a hint on how this test was written, would be great!

app/src/main/java/de/grobox/transportr/networks/TransportNetworkItem.java
ialokim: `TransportNetworkItem` extends `RegionItem` which itself extends `AbstractExpandableItem` setting at this point the `ViewHolder` class. To be able to override functions correctly, it has to be `RegionViewHolder` at this point.

But I'll look into it, I think I could use some anonymous typing here.

app/src/main/res/drawable/continent_africa.xml
ialokim: > Could you please simplify this massive path

How much more do I need to simplify? I've already simplified all them down to ~130 points.

> remove the translation group around it and use the proper heigth/width values already here

Could you please give me a hint on how to do this with Inkscape? 🤔

app/src/main/res/layout/list_item_transport_network.xml
ialokim: I'm not sure anymore, but I think this was producing different sizes which looks a bit odd. I added the value to keep a consistent text position throughout continent, country and transport network items.

How many PNG files are there? If there aren't too much, I'm ok with reverting this change here and open another PR converting the PNGs to vectorgraphics. Or at least resizing the PNGs to exactly 64 pixels width.

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
grote: This is an instrumentation test running on a real device, simulating user input. It is primarily used to take nice screenshots automatically. It is *not* run by the CI.

So you need to attach an Android device via ADB and run `All tests` in Android Studio.

app/src/main/res/drawable/continent_africa.xml
grote: In Inkscape, just select the group and ungroup it. in the status bar, it will tell you whether you selected a group or a path. Just ungroup the groups until they are gone.

app/src/main/res/layout/list_item_transport_network.xml
grote: Note that `dp` is not `pixels`. I wouldn't put work in enlarging PNGs. There is #238 and having the PNGs in a different size is a nice reminder of that.

app/build.gradle
ialokim: *`3.2.3` 😆

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
ialokim: Tests are now running, thanks to the continent ordering even simplier :)

app/src/main/java/de/grobox/transportr/networks/TransportNetworkItem.java
ialokim: I've now introduced type parameters, so this problem should be fixed.

app/src/main/res/drawable/continent_africa.xml
ialokim: Ok, the problem was the viewbox, I've removed it and simplified (a lot) the svg graphics. But I wasn't able to simplify more the actual path, within Inkscape it only lost the form but didn't reduce the number of points.

app/src/main/res/layout/list_item_transport_network.xml
ialokim: Ok, I've reverted this one.

ialokim:
Hereby I explicitly state that I accept the [Contributor License Agreement](/grote/transportr/blob/master/CLA.md) whenever contributing to Transportr.

ialokim:
I've now added some commits addressing your issues. There's still pending to add more translations for the continents, or will this be added by separate PRs in the future?

grote:
[clabot:check]

@ialokim thanks for the fixes! Before I look into them, would you mind looking into the lint errors that prevent the CI from succeeding?

grote:
@cla-bot[bot] check

ialokim:
> Before I look into them, would you mind looking into the lint errors that prevent the CI from succeeding?

Yes, there only were some unused translations that I removed now.

app/src/main/res/drawable/continent_africa.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="538.6851"
+    android:viewportHeight="645.0074"
+    android:width="538.68512dp"
+    android:height="645.00739dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_asia.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="950.9475"
+    android:viewportHeight="787.0276"
+    android:width="950.94751dp"
+    android:height="787.02759dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_central_america.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="136.3652"
+    android:viewportHeight="125.4169"
+    android:width="136.36523dp"
+    android:height="125.41689dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_europe.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="423.7945"
+    android:viewportHeight="337.6692"
+    android:width="423.79449dp"
+    android:height="337.66916dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_north_america.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="845.6545"
+    android:viewportHeight="565.9996"
+    android:width="845.65448dp"
+    android:height="565.99963dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_oceania.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="458.9524"
+    android:viewportHeight="454.8092"
+    android:width="458.95242dp"
+    android:height="454.80923dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/res/drawable/continent_south_america.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:viewportWidth="378.4156"
+    android:viewportHeight="610.44"
+    android:width="378.41562dp"
+    android:height="610.44dp">
```
grote: These should be `64dp` if I remember correctly. Bonus points for using `64px` in inkscape already and centering the drawing in it.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
```diff
@@ -79,25 +80,55 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+			for (Region subRegion : subRegions) {
+				if (subRegion instanceof Country) {
+					Country country = (Country)subRegion;
+					if (!regions.containsKey(country)) {
+						continue;
+					}
+					CountryItem countryItem = new CountryItem(country);
+					List<Region> networks = regions.get(country);
+					List<TransportNetworkItem> networkItems = new ArrayList<>(networks.size());
+					for (Region network : networks) {
```
grote: This triple-nested loop looks a bit crazy. I wonder if we can't simply things a bit so it is less complicated and easier to understand.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
```diff
@@ -146,19 +193,53 @@ private void selectItem() {
 			}
 		}
 	}
+	
+	private HashMap<Region, List<Region>> getRegionsByRegion() {
```
grote: `getRegionsByRegion()` sounds a bit confusing. Maybe we can find a clearer name?

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
grote: Maybe when the networks are defined, we could already define the hierarchy (best in Kotlin), so we save ourselves these iterations?

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
```diff
@@ -134,9 +165,25 @@ private void selectItem() {
...
+		Region region = network.getRegion();
+		Continent continent = null;
+		Country country = null;
+		if (region instanceof Country) {
+			country = (Country)region;
+			continent = country.getContinent();
+		} else if (region instanceof Continent) {
+			continent = (Continent)region;
+		}
+		pos = adapter.getPosition(new ContinentItem(continent));
```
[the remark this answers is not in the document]
ialokim: You're right, I've changed it.

app/src/main/java/de/grobox/transportr/networks/ContinentItem.java
```diff
@@ -0,0 +1,76 @@
...
+	private final Continent continent;
+
+	ContinentItem(Continent continent) {
+		super();
+		this.continent = continent;
+	}
+	
+	@Override
+	protected String getName(Context context) {
+		return this.continent.getName(context);
```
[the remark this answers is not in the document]
ialokim: Removed it.

app/src/main/java/de/grobox/transportr/networks/CountryItem.java
```diff
@@ -0,0 +1,76 @@
...
+	protected final Country country;
+
+	CountryItem(Country country) {
+		super();
+		this.country = country;
+	}
+
+	@Override
+	protected String getName(Context context) {
+		return this.country.getName(context);
```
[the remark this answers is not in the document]
ialokim: Removed it. Same for `TransportNetworkItem`.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
ialokim: > Maybe when the networks are defined, we could already define the hierarchy (best in Kotlin), so we save ourselves these iterations?

This could simplify it a lot, indeed. Where would be the right place for it then?

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
ialokim: could be `getSubRegionsByRegion`? But the functions wouldn't even be necessary when changing this whole iterations thing.

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
```diff
@@ -73,14 +73,9 @@ class PickTransportNetworkActivityTest : ScreengrabTest() {
                 .check(matches(withText(R.string.pick_network_first_run)))
         makeScreenshot("1_FirstStart")
 
-        // hack to find region position in list
-        val regionList = ArrayList(EnumSet.allOf(Region::class.java))
-        val context = InstrumentationRegistry.getTargetContext()
-        Collections.sort(regionList) { r1, r2 -> context.getString(r1.getName()).compareTo(context.getString(r2.getName())) }
-
```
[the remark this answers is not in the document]
ialokim: I could re-include the scrolling code if you like, searching for and opening Europe, then Germany and then Deutsche Bahn?

app/src/main/res/drawable/continent_africa.xml
ialokim: It wasn't necessary because I set the width in the list item definition. But I've changed it, what did I win with the bonus points? 😆

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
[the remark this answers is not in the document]
grote: Yes, I think that would be the safest thing to do.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
grote: We would probably do this right in the `TransportNetworks` interface. But maybe as a second step after this PR is done.

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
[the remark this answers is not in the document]
ialokim: Hmmm, while working on it, I thought it would be much easier to reimplement the scrolling function after changing the iterations method (probably as suggested in another PR?). For now, it would be necessary to copy the code from `PickTransportNetworkActivity` which I think is not the best option.

Btw, I've searched for this espresso testing functions and found [this documentation](https://developer.android.com/reference/android/support/test/espresso/contrib/RecyclerViewActions.html#actionOnItem(org.hamcrest.Matcher%3Candroid.view.View%3E,%20android.support.test.espresso.ViewAction)) which states:

> Performs a ViewAction on a view matched by viewHolderMatcher.
> 1. Scroll Recycler View to the view matched by itemViewMatcher
> 2. Perform an action on the matched view

So I don't quite understand why it doesn't work, I've tested it with another network which indeed was out of the display and it crashes. Would it be okay to handle this scrolling thing in the same PR suggested [here](https://github.com/grote/Transportr/pull/395#discussion_r168926895)?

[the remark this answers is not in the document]
grote: I would prefer not to merge anything which breaks tests.

grote:
@ialokim this has conflicts now and is still missing a fix for the tests.

app/src/androidTest/java/de/grobox/transportr/networks/PickTransportNetworkActivityTest.kt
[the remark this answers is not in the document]
ialokim: Fixed the test while introducing a `subRegions` list for `ParentRegion` (so, for `Continent` and `Country`). It's now using the same hack as before.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.java
ialokim: Removed this function in the latest commit.

ialokim:
The tests are now working again with the same hack as before and the conflicts because of the newly merged Nicaraguan provider are now also fixed.

grote:
Thanks @ialokim, I now merged this!

grote:
@ialokim I merged the improvements we talked about into master. Please have a look if everything is alright.

There still seems to be a bug with restoring the expansion and selection state when rotating the screen (best tested after clearing app data), but this seems to be an upstream bug.

ialokim:
> @ialokim I merged the improvements we talked about into master. Please have a look if everything is alright.

Are you refering to [this commit](https://github.com/grote/Transportr/commit/da2f9e8b567486b672f1bf768717cefae1de5c58)? I'm not very familiar with the Kotlin language, but looks all right for me. And indeed, much more simple! :+1: 

Great you're already moving towards RC for the 2.0 release! :tada:

PR #396, opened 2017-12-13
Improve np_desc_california text string.

n76:
Signed-off-by: Tod Fitch <Tod@FitchDesign.com>


grote:
Thanks for the change. Looks good to me! 

@Yahoe001 for you as well?

Issue #476, opened 2018-07-22
Location search for Helsinki region (HSL) does not work


aperezdc:
### Steps to reproduce

1. Install Transportr 2.0.0 using F-Droid (I haven't checked the Google Play version, I don't have Google Apps/Services installed).
2. Open the application, choose Finland→Helsinki (HSL) as region.
3. Tap the “Search here” entry and type some location, for example “Viiskulma” (it's an intersection), “Rautatientori” (a square) or “Keskuskatu” (a street).

### Expected outcome

The typed location can be chosen to navigate to.

### Actual outcome

Nothing happens, there are no search results.

### Additional information

- Trying to set a Home or Work address exhibits the same bahviour: no results are ever shown when typing the name of a location.
- Enabling ADB in the system settings and watching the output from `adb logcat` does not show anything particularly interesting.
- No errors are displayed in the UI either (e.g. no toast with an error message is produced).

aperezdc:
Let me know if there is some way in which I could help debugging this, I'll be happy to help :wink:

grote:
We should probably remove HSL, since they have shut down their API: https://github.com/schildbach/public-transport-enabler/issues/166

A pull request for removing it would be welcome!

If you want Finland supported, I suggest to [add a Navitia-based provider](https://transportr.app/contribute/#translating-into-your-language) or implement the new HSL API.

aperezdc:
@grote Thanks for the info, that is a bit sad... I *might* be able to hack something during the weekends, though—but bear with me, because my experience writing Android applications is close to zero. 

If I understand correctly, the GTFS data is available. After reading the contribution guide I have found that:

- https://maps.google.com/landing/transit/cities/index.html#Europe lists “Helsinging Seudun Liikene” (among others) for Finland.
- https://navitia.opendatasoft.com/explore/dataset/fi/ makes me think that Navitia has the data available.

I would like to know what would be the best approach to try to contribute updated support for HSL, and I have a few questions:

- Is the above enough to submit a Navitia based provider?
- Or would it be better to implement the [new HSL API](https://www.hsl.fi/en/opendata)?
- Would it be possible to use a Navitia based provider for the initial bringup, and then improve it with the additional information provided by the HSL API? For example, I see that the new HSL API has GTFS-Realtime feed for live traffic updates.
[...]

PR #479, opened 2018-07-24
Adapt to the updated HslProvider which now uses Navitia

aperezdc:
Related issue: #476 

----

Note that this requires https://github.com/schildbach/public-transport-enabler/pull/208 to be merged first.

Also, if I understand correctly, I'll have to update the PR later to bump the version of `public-transport-enabler` used in order to make the CI pass, right?

### DIFF
--- a/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
+++ b/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
@@ -407,11 +407,11 @@ private val networks = arrayOf(
             Country(
                 R.string.np_region_finland, flag = "🇫🇮", networks = listOf(
                     TransportNetwork(
-                        id = NetworkId.HSL,
+                        id = NetworkId.FI,
                         description = R.string.np_desc_hsl,
                         logo = R.drawable.network_hsl_logo,
                         status = BETA,
-                        factory = { HslProvider("pte_hsl", "Eixaeb9tnohcah7A") }
+                        factory = { HslProvider(NAVITIA) }
                     )
                 )
             ),


grote:
> Also, if I understand correctly, I'll have to update the PR later to bump the version of public-transport-enabler used in order to make the CI pass, right?

Correct, note that Transportr uses [the staging version of PTE](https://gitlab.com/opentransitmap/public-transport-enabler) where you can get your PR merged if it takes too long at PTE upstream.

aperezdc:
I accept the CLA.

aperezdc:
@grote Should I already post a PR to the PTE staging repository now, or is it better to wait a few days before doing that in case upstream picks the changes quickly enough?

grote:
I don't know. Up to you.

How well is this Navitia-based provider working compared to the previous one?

@cla-bot[bot] check

aperezdc:
I've checked a bit the results from Navitia using their API playground the other day, and for my daily commute and a few routes I take often the results are on par with those from Reittiopas (the HSL planner). The main differences I have seen so far are two:

- While Navitia will suggest bicycling sometimes, Reittiopas is smarter about this (when bicycling enabled), and will even suggest using city bikes if they are available. My guess is that the information about city bike stations is not exported to the data sets that Navitia uses. Probably because city bikes are only available for a few months around the summer, pickup stations are not fixed (they're in storage during winter), and the information about which ones have bikes is quite realtime.

- For some routes which are served both by train and bus, Navitia suggests using the bus a little bit more often than the train. From the look of it, the default settings of Navitia seems to favor less walking, and Reittiopas has a small bias towards trains and trams if the walking needed is not much more than using a bus. At any rate, even if the sorting of suggestions is a bit different, in the end both end up showing the same routes.

Personally, I don't think this is a problem, and the benefit of having the region supported again outweights those two minor issues.

grote:
Thanks for explaining! Nice to know the results are more or less the same!

> Is there any easy way I can tell Gradle to build Transportr with the local changes I have in PTE?

I suggest you use jitpack. This is already setup in Transportr. I include https://gitlab.com/opentransitmap/public-transport-enabler this way. So you would just need to change this dependency to point to your fork of PTE.

To update the dependency pinning, you need to remove the dependency hashes from build.gradle and then run `gradle -q calculateChecksums` to get the new version. Maybe you could add this information to https://github.com/grote/Transportr#building-from-source as well for others?

aperezdc:
@grote Thanks for the note about Jitpack, I managed to complete a debug build and it installed fine in my phone—I'll be testing it during the next days as well! At some point when I have another  couple of hours of free time I'll try to update the documentation as requested; probably in a couple of weeks.

aperezdc:
Now that I have a development build installed in my phone I got to notice the following: it seems that either the data fed to Navitia does not have detailed paths for the routes, in the screenshot one can clearly see that lines are straight between tram stops but they should be following the streets through which the line circulates. Or is this a known limitation of Navitia?

At any rate, I would still like to have this landed and keep the provider tagged as “beta” because it's already usable and better than no support at all, and use this as a motivation for me to reimplement `HslProvider` later using the official API. WDYT?

Straight lines from HSL route using Navitia

grote:
Both depends on the GTFS source file used by Navitia. I'd check this file if it defines the route colors and has a shape.text file. If not, you can complain with HSL about it.

If they are ignorant about the issue you could also ask Navitia if they are interested in applying the colors directly to their data. So you would need to maintain it in source code.

grote:
You PR landed upstream.

grote:
I made these two little changes myself now to be able to compile PTE again.

aperezdc:
@grote Yay, thanks! I am just back from holidays, and I'll try to find some time in the next days to keep improving things for the Helsinki region :+1:

PR #481, opened 2018-07-25
Add Pl Navitia provider (Municipal Transports of Poland)

mimi89999:

PR #451, opened 2018-06-02
Added Hungary provider

urbalazs:
Hi @grote 
This PR added Hungary provider, included two cities: Budapest (capital city) and Miskolc.
Related PR: https://github.com/schildbach/public-transport-enabler/pull/195

PR #486, opened 2018-07-27
Updated changelog

mimi89999:
I forgot to add the changelog in #481. I added it and the Hungary provider from #451.


grote:
Thanks! :D

PR #502, opened 2018-08-09
+eo to fastalane

verdulo:
aldonis Esperanton al fastlane-metadatumoj por vidigi en F-Droid
added Esperanto to fastlane-metadata to show in F-Droid

### DIFF
--- a/fastlane/generate_metadata.py
+++ b/fastlane/generate_metadata.py
@@ -17,6 +17,7 @@ LANG_MAP = {
     'values-cs': 'cs-CZ',
     'values-de': 'de-DE',
     'values-el': 'el-GR',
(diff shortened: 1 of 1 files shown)


grote:
Did you check if this is supported by Google Play as well? Because if not, uploading the app to Google Play will fail.

verdulo:
It is probably not supported, but please at least try. Why this discussion is in English? When people will stop use this stupid language for international communication?

grote:
I am getting ```/home/user/.gem/ruby/gems/google-api-client-0.21.2/lib/google/apis/core/http_command.rb:218:in `check_status': [!] unsupportedListingLanguage: Listing language eo is not supported. (Google::Apis::ClientError)```

PR #535, opened 2018-10-18
Show swap locations button

ByteHamster:
Since the design was updated, I constantly keep entering the "From" location into the search bar on the main page. Therefore, after pressing the routing icon, the location appears in the wrong field. I need to use the "Swap locations" button a lot since the update.

This PR shows the button in the toolbar (if room) for faster access.


ByteHamster:

grote:
Thanks for the PR @ByteHamster!

@cla-bot[bot] check

app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.java
```diff
@@ -197,6 +199,12 @@ public void onCreateOptionsMenu(Menu menu, MenuInflater inflater) {
 					})
 					.show();
 		}
+
+		Drawable drawable = menu.findItem(R.id.action_swap_locations).getIcon();
+		if (drawable != null) {
+			drawable.mutate();
+			drawable.setColorFilter(0xffffffff, PorterDuff.Mode.SRC_ATOP);
+		}
```
grote: I think we might not be using this icon elsewhere, so we could just set it to white in the XML.

ByteHamster: It is used when pressing the overflow menu for favorites

grote: Right, so we need to mutate here.

PR #537, opened 2018-10-20
Canadian networks (Ontario and Quebec) are given proportional flags

anjandev:
The unmodified flags of Ontario and Quebec (svgs downloaded from wikipedia). Completely to proportion.

They also look better than my flags from before:

screenshot_20181019-212614_transportr_devel
screenshot_20181019-212601_transportr_devel


grote:
I kind of prefer the maple leaf we have now. It also looks better in the circle view of the navigation drawer. The non-quadratic flag leaves space at the top and bottom.

Altonss:
> I kind of prefer the maple leaf we have now. It also looks better in the circle view of the navigation drawer. The non-quadratic flag leaves space at the top and bottom.

It's a matter of taste, but if we want to leave it as it is maybe the PR can be closed?

ialokim:
Closing for now, as long as nobody wants to pick it up again.

PR #542, opened 2018-10-31
Add notification support on DeparturesActivity for e.g. smartwatches

ikarulus:
Getting something countdown-like for the next departures per station on my pebble watch was always on my wishlist for my public transit "workflow". I used transportr before, so I forked this for a hackathon and just did it. Maybe it's useful for the main app.
Using notifications here seemed like a good idea, because that's vendor agnostic and even not completely useless without such a watch. The notifications should update every 30s. Notifications can be toggled for 10 Minutes with a menu button in the DeparturesActivity.
img_20181027_131651 1


ikarulus:

PR #576, opened 2019-01-26
Fix trip delays

### LINKED ISSUE #575: Double Minus and Wrong Color of Negative Delay
There should only be one minus and it should be green. The color is correct in the TripDetails screen.

screenshot_20190124-172914_transportr_devel

ialokim:
This PR unifies the delay handling for `trips.search` with `trips.detail` and by doing so applies the right color to trips with "negative delay" (so public transport services ahead of their time). It also fixes the bug to not display a double minus in front of such delays.

The second commit just moves `BaseViewHolder` to the `trips` package as it is now used by both subpackages.

Fixes #575

### DIFF
--- a/app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
+++ b/app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
@@ -17,7 +17,7 @@
  *    along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package de.grobox.transportr.trips.detail
+package de.grobox.transportr.trips
 
 import android.content.Context
 import android.view.View
@@ -33,7 +33,7 @@ import de.schildbach.pte.dto.Position
 import de.schildbach.pte.dto.Stop
 import java.util.*
 
-internal abstract class BaseViewHolder(v: View, protected val listener: LegClickListener) : RecyclerView.ViewHolder(v) {
+internal abstract class BaseViewHolder(v: View) : RecyclerView.ViewHolder(v) {
 
     protected val context: Context = v.context
     protected val fromTime: TextView = v.findViewById(R.id.fromTime)
@@ -41,7 +41,7 @@ internal abstract class BaseViewHolder(v: View, protected val listener: LegClick
     protected val fromDelay: TextView = v.findViewById(R.id.fromDelay)
     protected val toDelay: TextView = v.findViewById(R.id.toDelay)
 
-    fun setArrivalTimes(timeView: TextView, delayView: TextView, stop: Stop) {
+    fun setArrivalTimes(timeView: TextView?, delayView: TextView, stop: Stop) {
         if (stop.arrivalTime == null) return
 
         val time = Date(stop.arrivalTime.time)
@@ -58,10 +58,10 @@ internal abstract class BaseViewHolder(v: View, protected val listener: LegClick
         } else {
             delayView.visibility = GONE
         }
-        timeView.text = getTime(context, time)
+        timeView?.let { it.text = getTime(context, time) }
     }
 
-    fun setDepartureTimes(timeView: TextView, delayView: TextView, stop: Stop) {
+    fun setDepartureTimes(timeView: TextView?, delayView: TextView, stop: Stop) {
         if (stop.departureTime == null) return
 
         val time = Date(stop.departureTime.time)
@@ -78,7 +78,7 @@ internal abstract class BaseViewHolder(v: View, protected val listener: LegClick
         } else {
             delayView.visibility = GONE
         }
-        timeView.text = getTime(context, time)
+        timeView?.let { it.text = getTime(context, time) }
     }
 
     protected fun TextView.addPlatform(position: Position?) {
--- a/app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
+++ b/app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
@@ -33,6 +33,7 @@ import androidx.recyclerview.widget.RecyclerView
 import com.google.common.base.Strings
 import com.google.common.base.Strings.isNullOrEmpty
 import de.grobox.transportr.R
+import de.grobox.transportr.trips.BaseViewHolder
 import de.grobox.transportr.trips.detail.LegViewHolder.LegType.*
 import de.grobox.transportr.ui.LineView
 import de.grobox.transportr.utils.DateUtils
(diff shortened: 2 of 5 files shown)


app/src/main/java/de/grobox/transportr/utils/DateUtils.java
```diff
@@ -83,13 +83,7 @@ public static String getDelayString(long delay) {
...
 	@Nullable
 	public static String getDelayText(long delay) {
-		if (delay >= 0) {
-			return "+" + Long.toString(delay / 1000 / 60);
-		} else if (delay < 0) {
-			return "-" + Long.toString(delay / 1000 / 60);
-		} else {
-			return null;
-		}
+		return (delay >= 0 ? "+" : "") + Long.toString(delay / 1000 / 60);
```
grote: I think this isn't `@Nullable` anymore, right? Could you please remove the annotation and check all usages of this function please and remove `null` handling there?

app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
```diff
@@ -33,20 +33,20 @@ import de.schildbach.pte.dto.Position
...
+internal abstract class BaseViewHolder(v: View) : RecyclerView.ViewHolder(v) {
 
     protected val context: Context = v.context
     protected val fromTime: TextView = v.findViewById(R.id.fromTime)
     protected val toTime: TextView = v.findViewById(R.id.toTime)
     protected val fromDelay: TextView = v.findViewById(R.id.fromDelay)
     protected val toDelay: TextView = v.findViewById(R.id.toDelay)
 
-    fun setArrivalTimes(timeView: TextView, delayView: TextView, stop: Stop) {
+    fun setArrivalTimes(timeView: TextView?, delayView: TextView, stop: Stop, showZeroDelay: Boolean = true) {
```
grote: Why do we need `showZeroDelay`? I think the convention was (or should be) to show a green +0 if real time information is available and to show nothing if it isn't available.

app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
```diff
@@ -26,12 +26,13 @@ import android.view.View.VISIBLE
...
 import android.widget.ImageView
 import android.widget.TextView
+import de.grobox.transportr.trips.BaseViewHolder
 import de.grobox.transportr.utils.TransportrUtils.getLocationName
 import de.schildbach.pte.dto.Stop
 import kotlinx.android.synthetic.main.list_item_stop.view.*
 
 
-internal class StopViewHolder(v: View, listener: LegClickListener) : BaseViewHolder(v, listener) {
+internal class StopViewHolder(v: View, val listener : LegClickListener) : BaseViewHolder(v) {
```
grote: Could the listener be `private`?

app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
```diff
@@ -45,7 +46,7 @@ import de.schildbach.pte.dto.Trip.*
 import kotlinx.android.synthetic.main.list_item_leg.view.*
 
 
-internal class LegViewHolder(v: View, listener: LegClickListener, private val showLineName: Boolean) : BaseViewHolder(v, listener) {
+internal class LegViewHolder(v: View, protected val listener: LegClickListener, private val showLineName: Boolean) : BaseViewHolder(v) {
```
grote: Could the listener be `private`?

app/src/main/java/de/grobox/transportr/trips/search/TripViewHolder.kt
```diff
@@ -36,39 +36,29 @@ import de.schildbach.pte.dto.Trip
...
 
     fun bind(trip: Trip, listener: OnTripClickListener) {
         // Relative Departure Time
         setRelativeDepartureTime(fromTimeRel, trip.firstDepartureTime)
 
-        // Departure Time
+        // Departure Time and Delay
         val firstLeg = trip.legs[0]
         if (firstLeg is Public) {
-            fromTime.text = getTime(context, firstLeg.getDepartureTime(true))
```
grote: Did you check that what we do in `setDepartureTimes()` is equivalent to what happens in `firstLeg.getDepartureTime(true)`?

app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
ialokim: Up to now, the green `+0` is only shown in the trip details view, not in the search results view (if real time information is available, obviously). So I have added this flag to be able to differentiate between the two cases. But if you want to change the app's behavior here, it would be easy to adopt.

app/src/main/java/de/grobox/transportr/trips/search/TripViewHolder.kt
ialokim: It is equivalent to what happened before in this line and in lines 66-71.

app/src/main/java/de/grobox/transportr/utils/DateUtils.java
ialokim: True, I have removed the annotation, but the only two usages were not handling the `null` case anyway (as I have been changing them within this PR).

app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
ialokim: Indeed, I have changed it.

app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
ialokim: Indeed, I have changed it.

ialokim:
Thanks a lot for your review! Do you want me to add another commit applying the three (little) changes or do you prefer rebasing and force-pushing for cleaner git history?

app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
grote: I think we should indicate the availability of real-time information everywhere consistently. So we could drop the flag, right?

grote:
Thanks for the fixes! Would be nice to force push all three commits into a single one.

app/src/main/java/de/grobox/transportr/trips/BaseViewHolder.kt
ialokim: So then yes, the flag can be dropped.

ialokim:
> Would be nice to force push all three commits into a single one.

Okay, done. I have applied all the changes and force-pushed the new version.

grote:
Awesome, thanks!

PR #584, opened 2019-02-13
Don't search with via location if via field is hidden

### LINKED ISSUE #522: Deactivated "Via" field still triggers error
**Description**
If the "Via" field is equal to the "From" or "To" field an error message is displayed.
As far as good. But this error is also displayed when "Via" is deactivated.

**To Reproduce**
1. Search for an Connection (also add a place to the "Via" field)
2. Deactivate the Via field
3. Use the "From" or "To" field and add the place which have been in the "Via" field before deactivating it

**Expected behavior**
The error message "The 'from' and 'to' location are too close together" appears.
This happens even if the two places are several hundred kilometers separated from each other.

**Screenshot**
photo_2018-09-08_19-41-20

**Version**
 - **Transportr Version:** [2.0.2]
 - **Device:** [HTC one M8s]
 - **Android Version:** [6.0.1]

ialokim:
This PR addresses #522 and solves even a greater issue: If a via location has been chosen before hiding the via location field, the via location is still used to perform the query.

Unfortunately, Android Studio reports an "Unboxing of `isExpanded.getValue()` may produce a NullPointerException" warning, the only way to avoid this is to check manually for `null` type (which should never happen btw, as `isExpanded` is instantiated when instantiating the `ViewModel`). Not sure what would be the way to go in this case.

Fixes #522

### DIFF
--- a/app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.java
+++ b/app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.java
@@ -224,7 +224,11 @@ public class DirectionsViewModel extends SavedSearchesViewModel implements TimeD
 			if (calendar == null) return;
 		}
 
-		TripQuery tripQuery = new TripQuery(fromLocation.getValue(), viaLocation.getValue(), toLocation.getValue(),
+		WrapLocation via = null;
+		if (isExpanded.getValue() != null && isExpanded.getValue())
+			via = viaLocation.getValue();
+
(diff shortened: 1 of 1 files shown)


grote:
If you are really really sure the value can't be null, you can enclose the get method in requireNonNull()

ialokim:
To avoid any NullPointerExceptions at runtime, I've added the `!= null` check now.

grote:
Thanks!

PR #588, opened 2019-02-17
Align stops in TripDetailsFragment even without times

### LINKED ISSUE #587: Stops without times not indented
**Describe the bug**
The Swiss SBB :switzerland: includes points of interests (like the Gotthard Base Tunnel) as "stops" in their train connections.
Since howerver they are missing arrival/departure times, Transportr renders them shifted to the left and they become malaligned. It's best explained by the screenshots below.

**To Reproduce**
1. Choose Siwtzerland -> SBB as a network provider
2. Choose an affected train connection, e.g: Zurich HB -> Milano Centrale
3. Select any connection
4. Expand all stops

**Expected behavior**
All names should be indented onto the same line.

**Screenshots**

Zurich - Milano | Zurich - Brig | Zurich - Chur
---|---|---
img_20190216_152001 | img_20190216_151921 | img_20190216_151941

**Versions:**
 - **Transportr Version:** [2.0.5]
 - **Device:** [irrelevant]
 - **Android Version:** [7.1.2]

ialokim:
When no `arrivalTime` is set, the width of the `fromTime` TextView is changed to the width of the formatted current time, preventing aligning errors.

Fixes #587

### DIFF
--- a/app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
+++ b/app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
@@ -21,15 +21,16 @@ package de.grobox.transportr.trips.detail
 
 
 import android.view.View
-import android.view.View.GONE
-import android.view.View.VISIBLE
+import android.view.View.*
 import android.widget.ImageButton
 import android.widget.ImageView
 import android.widget.TextView
 import de.grobox.transportr.trips.BaseViewHolder
+import de.grobox.transportr.utils.DateUtils.getTime
 import de.grobox.transportr.utils.TransportrUtils.getLocationName
 import de.schildbach.pte.dto.Stop
 import kotlinx.android.synthetic.main.list_item_stop.view.*
+import java.util.Date
 
 
 internal class StopViewHolder(v: View, private val listener : LegClickListener) : BaseViewHolder(v) {
@@ -41,8 +42,16 @@ internal class StopViewHolder(v: View, private val listener : LegClickListener) 
     fun bind(stop: Stop, color: Int) {
         if (stop.arrivalTime != null) {
             setArrivalTimes(fromTime, fromDelay, stop)
+            fromTime.visibility = VISIBLE
         } else {
             fromDelay.visibility = GONE
+            if (stop.departureTime == null) {
+                // insert dummy time field for stops without times set, so that stop circles align
+                fromTime.text = getTime(context, Date())
+                fromTime.visibility = INVISIBLE
+            } else {
+                fromTime.visibility = GONE
+            }
         }
 
         if (stop.departureTime != null) {


app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
```diff
@@ -42,6 +43,9 @@ internal class StopViewHolder(v: View, private val listener : LegClickListener)
         if (stop.arrivalTime != null) {
             setArrivalTimes(fromTime, fromDelay, stop)
         } else {
+            // ensure fields for stops without times have a default width, so that stop circles align
+            fromTime.minWidth = fromTime.paint.measureText(getTime(context, Date())).toInt()
```
grote: This feels a bit hacky. I feel like there should be a ConstraintLayout solution for this.

ialokim: True, feels odd, but I didn't find any solution with ContraintLayout: I think you can not reference a view outside the current list item and if there are no times given, you can't guess the width used in other list items. `minWidth` isn't an option neither as the width depends heavily on the time format ("10:00 PM" is much larger than "10:00" for example).

But if you find any other solution, that would be great!

ialokim: An alternative would be to set a dummy time in this case and set the TextView to invisible, perhaps a little less hacky.

grote: Right, I had forgotten about the problem of different line items. The invisible dummy time sounds a bit better then the measure approach.

Just make sure the `fromTime` is set to visible when we have a stop time, otherwise it will disappear when the line item view gets recycled with a stop time.

ialokim:
I've switched to the invisible dummy time and added a line setting `fromTime` to `VISIBLE` again in case it does have a time to display.

app/src/main/java/de/grobox/transportr/trips/detail/StopViewHolder.kt
```diff
@@ -41,8 +42,14 @@ internal class StopViewHolder(v: View, private val listener : LegClickListener)
...
         if (stop.arrivalTime != null) {
             setArrivalTimes(fromTime, fromDelay, stop)
+            fromTime.visibility = VISIBLE
         } else {
             fromDelay.visibility = GONE
+            if (stop.departureTime == null) {
+                // insert dummy time field for stops without times set, so that stop circles align
+                fromTime.text = getTime(context, Date())
+                fromTime.visibility = INVISIBLE
+            }
```
grote: I think there's still a case where a recycled view holder will be used with `fromTime` invisible: `stop.arrivalTime == null && stop.departureTime != null`.

ialokim: Okay, but when `stop.arrivalTime == null` then we actually don't need `fromTime` right? If `stop.departureTime != null` the *`toTime`* will be set some lines later and changed to be visible.

grote: I think `stop.departureTime` is unrelated, right?

I am concerned that even if we don't "need" `fromTime` when there's no arrival time, that its visibility is undefined in that case. It might be visible or it might not be. We would need to test both cases to see how this looks in the UI.

ialokim: Hum, the current behavior is that `fromTime` is always visible whether or not it is set to `stop.arrivalTime`. So a visible but unused `fromTime` doesn't harm (as it is now) and an invisible `fromTime` either (tested in Swiss case).

But you might be right, I think instead of hoping it will look good in the UI, I could simply add another `fromTime.visibility = VISIBLE` in a new `else` case in line 52.

grote: You could ;)

grote: Right. Would you mind testing these cases? At least in the graphic LayoutEditor?

ialokim: Sadly, it doesn't display as it should for me in the Android Studio XML preview even without changing anything. Any ideas why this is happening?

transportr

ialokim: I managed to test it now after `Build > Clean Project` in the XML preview with the different visibility settings. Should be ready to merge now.

PR #592, opened 2019-03-04
Add Lisbon

brandsimon:
I added lisbon.
If there is any problem. I am happy to adapt the PR.
If this PR got accepted, when will be the next rebuild on F-Droid?

### DIFF
--- a/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
+++ b/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
@@ -512,6 +512,19 @@ private val networks = arrayOf(
                 )
             ),
             Country(
+                R.string.np_region_pt, flag = "🇵🇹", networks = listOf(
+                    TransportNetwork(
+                        id = NetworkId.PT,
+                        name = R.string.np_name_pt,
+                        description = R.string.np_desc_pt,
+                        agencies = R.string.np_desc_pt_networks,
+                        logo = R.drawable.network_portugal_logo,
+                        status = BETA,
+                        factory = { PortugalProvider(NAVITIA) }
+                    )
+                )
+            ),
+            Country(
                 R.string.np_name_spain, flag = "🇪🇸", networks = listOf(
                     TransportNetwork(
                         id = NetworkId.SPAIN,
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -228,6 +228,11 @@ and always knows where you are to not miss where to get off the bus.
 	<string name="np_name_sydney" translatable="false">Transport NSW</string>
 	<string name="np_desc_sydney">Australia, New South Wales, Sydney</string>
 
+	<string name="np_region_pt">Portugal</string>
+	<string name="np_name_pt">Portugal</string>
+	<string name="np_desc_pt">Lisbon</string>
+	<string name="np_desc_pt_networks">Bus, Train, Funicular, Ferry, Metro</string>
+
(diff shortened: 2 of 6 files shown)


brandsimon:
I agree to the Contributor License Agreement for all PRs to this repo.

grote:
Thanks a lot for your contribution!

@cla-bot[bot] check

app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
```diff
@@ -511,6 +511,18 @@ private val networks = arrayOf(
                     )
                 )
             ),
+            Country(
+                R.string.np_name_pt, flag = "🇵🇹", networks = listOf(
```
ialokim: I think this should be `R.string.np_region_pt` as it is the country name and not the network's one.

app/src/main/res/values/strings.xml
```diff
@@ -228,6 +228,11 @@ and always knows where you are to not miss where to get off the bus.
 	<string name="np_name_sydney" translatable="false">Transport NSW</string>
 	<string name="np_desc_sydney">Australia, New South Wales, Sydney</string>
 
+	<string name="np_region_pt">Portugal</string>
+	<string name="np_name_pt">Portugal</string>
```
ialokim: I would opt for using *Lisbon* as provider name here (if it only provides information for this city?). What's your opinion here @grote?

grote: I think this is Navitia's Portugal region. If it only supports Lisbon at the moment, this could be used instead. However, then it needs to be changed when they support other cities.

app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
brandsimon: @ialokim Yes, I agree, done.

brandsimon:
@ialokim @grote I added the logo. I will squash before merge.

app/src/main/res/values/strings.xml
ialokim: Okay, then we leave it as it is.

ialokim:
Nice, thanks for adding the logo! Could you also add the SVG you have used before conversion to the `artworks/networks` folder as described [here](https://transportr.app/contribute/#transport-network-logos)?

brandsimon:
@ialokim Done and squashed.
What is the svg file used for?

artwork/networks/network_portugal_logo.svg
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="25" height="64">
```
ialokim: Shouldn't it be `width="64"` here?

ialokim:
> What is the svg file used for?

Not sure, I think it is just for reference. Could you squash all commits regarding the provider together (so, except the PTE version bump)?

artwork/networks/network_portugal_logo.svg
brandsimon: Yes, sorry, I should have looked at it again :-/

brandsimon:
@ialokim Done.

ialokim:
Great, thanks for your patience! I've tested the new provider and it should be ready to be merged now @grote.

grote:
Thanks @brandsimon and @ialokim!

PR #615, opened 2019-06-19
Don't post screenshots

schildbach:


schildbach:
I'm afraid I don't sign CLAs.

grote:
I don't think a CLA is needed for such a small change, but I am worried that users might not know what a stacktrace is. Maybe better say "copy the text above"?

People also complain about the URL being wrong. What is the right one that should be used?

schildbach:
I'd say `https://github.com/schildbach/public-transport-enabler/issues` is what should be linked to.

TheLastProject:
Why not just remove the instructions and add a title and body GET param to the new issue URL so it gets prefilled with the correct data? https://help.github.com/en/articles/about-automation-for-issues-and-pull-requests-with-query-parameters

ialokim:
This has been resolved with #625.

PR #627, opened 2019-08-01
Display price information

ialokim:
This PR addresses #231 by adding standard ticket price information (picking the `Fare.Type.ADULT` one as standard) to the trip search results and details view. It mainly follows the mock-ups and the discussion at #231. I didn't add an color value animation to switch between red and white background as it would not fit well into the current setting where the toolbar and the bottomBar are displayed only after the swipe gesture being finished.

I've also switched `TripDetailFragment` from Java to Kotlin code to take advantage of Kotlin extensions.

The code has been tested with both themes on a local Android emulator.


app/src/main/java/de/grobox/transportr/trips/detail/TripDetailFragment.kt
```diff
@@ -0,0 +1,203 @@
...
+        fromTime = bottomBar.findViewById(R.id.fromTime)
+        from = bottomBar.findViewById(R.id.from)
+        toTime = bottomBar.findViewById(R.id.toTime)
+        to = bottomBar.findViewById(R.id.to)
+        duration = bottomBar.findViewById(R.id.duration)
+        price = bottomBar.findViewById(R.id.price)
+        topBar = v.findViewById(R.id.topBar)
+        fromTimeRel = topBar.findViewById(R.id.fromTimeRel)
+        durationTop = topBar.findViewById(R.id.durationTop)
+        priceTop = topBar.findViewById(R.id.priceTop)
```
grote: You should be able to remove those and use the kotlin databinding instead

app/src/main/res/layout/fragment_trip_detail.xml
```diff
@@ -87,24 +89,105 @@
...
+				android:layout_height="match_parent"
+				android:orientation="vertical"
+				android:layout_marginTop="8dp"
+				android:layout_marginEnd="8dp"
+				android:layout_marginRight="8dp"
+				android:layout_marginBottom="8dp"
+				android:gravity="center_vertical"
+				app:layout_constraintEnd_toEndOf="parent"
+				app:layout_constraintTop_toTopOf="parent"
+				app:layout_constraintBottom_toBottomOf="parent">
```
grote: Would be nice to not introduce another hierarchy here, but use the existing ConstraintLayout for layouting.

app/src/main/res/layout/fragment_trip_detail.xml
```diff
@@ -87,24 +89,105 @@
...
+
+		</androidx.constraintlayout.widget.ConstraintLayout>
+
+		<androidx.constraintlayout.widget.ConstraintLayout
+			android:id="@+id/topBar"
+			android:layout_width="match_parent"
+			android:layout_height="wrap_content"
+			android:layout_marginTop="0dp"
+			android:background="?android:colorBackground"
+			tools:visibility="visible">
```
grote: Why a new ConstraintLayout?

app/src/main/res/layout/list_item_trip.xml
```diff
@@ -87,24 +87,41 @@
...
-			android:layout_marginBottom="8dp"
-			android:layout_marginTop="8dp"
-			app:layout_constraintBottom_toBottomOf="@+id/lines"
-			app:layout_constraintRight_toRightOf="parent"
+			android:orientation="vertical"
+			android:gravity="center_vertical"
+			app:layout_constraintEnd_toEndOf="parent"
 			app:layout_constraintTop_toTopOf="@+id/lines"
-			tools:text="1:30"/>
+			app:layout_constraintBottom_toBottomOf="@+id/lines">
```
grote: Would be nice to not introduce another hierarchy here, but use the existing ConstraintLayout for layouting.

app/src/main/java/de/grobox/transportr/trips/detail/TripDetailFragment.kt
ialokim: I'm not sure if I understand your proposal: Do you mean by just removing the `.findViewById(R.id.` part or is it even possible to remove all of these lines and initialize them somehow right away starting from line 68?

app/src/main/res/layout/fragment_trip_detail.xml
ialokim: I'm not sure if that is really feasible because the price might not be shown if the information is not available. At least in my tries I didn't manage to layout it nicely in both cases with and without price.

app/src/main/res/layout/fragment_trip_detail.xml
ialokim: Because the two ConstraintLayouts are not shown at the same time, either the `bottomBar` or the `topBar` is.

app/src/main/res/layout/list_item_trip.xml
ialokim: Here I've encountered the same problem as described above with layouting for both cases with and without price information.

app/src/main/java/de/grobox/transportr/trips/detail/TripDetailFragment.kt
grote: You should be able to use Kotlin ViewBinding https://kotlinlang.org/docs/tutorials/android-plugin.html#view-binding

app/src/main/res/layout/fragment_trip_detail.xml
grote: What issue specifically did you have? Note there's other views that get hidden and shown even dynamically. The ConstraintLayout should give you everything you need for avoiding nested layouts here.

app/src/main/res/layout/fragment_trip_detail.xml
grote: Sorry can you explain a bit more?

ialokim: Sure, `bottomBar` is the what's visible if the map is spanning almost all of the screen and the start and end points of the journey are just shown in the bottom line (with red background). Whereas `topBar` is the newly added line above the journey details as proposed in https://github.com/grote/Transportr/issues/231#issuecomment-355644068.

This means both are never visible at the same time and are thus completely independent one from another. In [TripDetailFragment.kt](https://github.com/grote/Transportr/pull/627/files#diff-112f199a9477cb8f84092554d964f660R169), the visibility is switched for both ConstraintLayouts whenever the sheet state has changed. That's why I went with having two separate layout "groups" which can be easily shown or hidden altogether.

app/src/main/java/de/grobox/transportr/trips/detail/TripDetailFragment.kt
ialokim: Okay, nice hint, this is really useful!

app/src/main/res/layout/fragment_trip_detail.xml
ialokim: Thanks for insisting, after some searching in the android docs I found ConstraintLayout Barriers to be quite useful for these kind of problem. Should be fixed with the last commit.

app/src/main/res/layout/list_item_trip.xml
ialokim: See above.

ialokim:
I've incorporated your change requests by using Kotlin ViewBinding and ConstraintLayout Barriers instead of an inner LayoutGroup. I've also changed some extra constraints in one of the layouts for using a second Barrier at the left side too.

As mentioned in #632 I had to fix some gradle dependency verification issues after rebasing onto current master, that's the first of the new commits. Feel free to use it or not, if it's not necessary.

app/src/main/java/de/grobox/transportr/trips/detail/TripDetailFragment.kt
```diff
@@ -0,0 +1,178 @@
...
+        viewModel.sheetState.value = MIDDLE
+    }
+
+    private fun onSheetStateChanged(sheetState: SheetState?) {
+        when (sheetState) {
+            null -> return
+            BOTTOM -> {
+                toolbar.visibility = GONE
+                topBar.visibility = GONE
+                bottomBar.visibility = VISIBLE
```
grote: These visibility changes appear a bit rough in the UI. Should be easy to use a object animator instead here to make things more smooth. But we can do this in a separate MR.

app/src/main/java/de/grobox/transportr/trips/detail/TripUtils.kt
```diff
@@ -135,4 +137,18 @@ internal object TripUtils {
...
+    }
+
+    fun Trip.getStandardFare(): String? {
+        fares?.find { fare -> fare.type == Fare.Type.ADULT }?.let {
+            val format = NumberFormat.getCurrencyInstance()
+            format.currency = it.currency
+            return format.format(it.fare)
+        }
+        return null
+    }
```
grote: Nice!

PR #632, opened 2019-10-15
adapt witness.gradle to build successfully

ialokim:
While rebasing onto current master for #627 I encountered some verification issues with gradle dependencies. I'm not really sure whether this is a weird bug due to my local system, but at least for me after applying these changes it built successfully again.


ialokim:
I did some research and concluded the problem indeed lays in some different configuration, comparable to what has been described [here](https://code.briarproject.org/briar/briar/issues/1488). I guess the problem is that I'm using Java 11 and the CI is using Java 8.

Thus closing this issue for now, but perhaps we should mention that a specific platform and configuration is needed for a successful build.

grote:
@ialokim example for specifying a gradle wrapper hash: https://code.briarproject.org/briar/briar/blob/master/gradle/wrapper/gradle-wrapper.properties#L6

PR #636, opened 2019-10-18
Add gradle wrapper and adapt witness.gradle to build successfully

ialokim:
I've also changed the `update-dependency-pinning.sh` script to use the gradle wrapper and filter out environment-dependent hashsums.


update-dependency-pinning.sh
```diff
@@ -3,5 +3,5 @@
 WITNESS=app/witness.gradle
 
 echo "" > $WITNESS
-gradle -q calculateChecksums | grep -v "^\(Skipping\|Verifying\|        'transforms-1\)" | grep -v transforms-2 > $WITNESS
+./gradlew -q calculateChecksums | grep -Ev "^(Skipping|Verifying)" | grep -Ev "files-2.1:|:build-tools:|:platforms:" > $WITNESS
```
grote: Can we maybe make this more narrow, so we don't exclude anything by mistake and don't notice?

Maybe noWitness can also do more? https://github.com/grote/Transportr/blob/master/gradle.properties#L3

grote:
Thanks, looks great!

update-dependency-pinning.sh
ialokim: I couldn't find anything documented on this `noWitness` property. Where is the source code for the `gradle-witness.jar` used in Transportr?

ialokim: I guess it's this one: https://github.com/akwizgran/gradle-witness?

Unfortunately, `noWitness` didn't work for me as the problematic dependencies are part of various configurations. Even using `printDependencies` as mentioned in walleth/walleth#354 and excluding all configurations having `files-2.1` as dependency with `noWitness` did not work out for some reason.

The only thing I could think of to make things safer would be using `:build-tools:core-lambda-stubs.jar:` and `:platforms:android.jar:` in the script instead of the shorter regex above.

grote: Yeah better have the larger regexes in there.

ialokim: I've changed it accordingly.

grote: Thanks!

grote:
I forgot that the README also needs to be updated: https://github.com/grote/Transportr#building-from-source

PR #642, opened 2019-11-02
adapt README to use gradle wrapper

ialokim:
This PR adapts the README after switching to gradle wrapper for this project as requested by @grote in https://github.com/grote/Transportr/pull/636#issuecomment-548798473.


grote:
Thanks for the quick fix!

PR #656, opened 2019-12-28
Departures screen: Always show realtime information

ialokim:
If realtime information is available, mark departures as such by displaying a green "+0" so that one can distinguish between on-time transports and ones without realtime information.

This PR also migrates `DepartureViewHolder` from Java to Kotlin.


app/src/main/java/de/grobox/transportr/departures/DepartureViewHolder.kt
```diff
@@ -0,0 +1,105 @@
...
+        setRelativeDepartureTime(timeRel, predictedTime ?: plannedTime)
+        timeAbs.text = getTime(timeAbs.context, plannedTime)
+        predictedTime?.let {
+            val delayTime = it.time - plannedTime.time
+            delay.text = getDelayText(delayTime)
+            if (delayTime <= 0)
+                delay.setTextColor(ContextCompat.getColor(delay.context, R.color.md_green_500))
+            else
+                delay.setTextColor(ContextCompat.getColor(delay.context, R.color.md_red_500))
+            delay.visibility = VISIBLE
```
grote: Wasn't there a helper function for this? I think trips do the same and have the same logic, right?

ialokim: It's not exactly the same as PTE returns quite different data for trips and departures, e.g. delay is returned as field for trips but not for departures.

Thus only these four last lines could be moved to a new helper function.

grote:
Thanks for already migrating more classes to Kotlin. That's great! Would you mind doing the migration in a separate commit in the future, so I can review migration and change separately?

app/src/main/java/de/grobox/transportr/departures/DepartureViewHolder.kt
grote: Ah cool, well if you think it makes sense to factor these bits out feel free to, otherwise it can stay as is as well.

ialokim: I would leave it as is for now.

ialokim:
> migration in a separate commit in the future

Sure, no problem. I'll try to remember :smile:

PR #657, opened 2019-12-28
Departures: Show error if returned list is empty

### LINKED ISSUE #623: [Crash] [VVT TNetwork] Departures
When touching on Departures (of a station) the up-coming site is blank and if you then swiping down the app crashes. This issue just appears to me if I use the VVT Transport Network.

ialokim:
This PR partly fixes #623 by checking if PTE returns an empty list. In this case, the error page is shown.


grote:
Cool, why is this only fixing the issue partly?

ialokim:
Well the other issue is that the returned list is indeed empty in some networks, but that should be an upstream issue then.

grote:
But it still shouldn't crash in that case, but show a message that no departures could be found, right?

ialokim:
Exactly, that's how it behaves with this PR.

PR #658, opened 2019-12-28
Home and Work: Add popup menu entry to get departures

ialokim:
Often it is useful to get departures at a station right away, particularly at the home or work location. This PR replaces the "Edit Search" option in the corresponding popup-menu (which is not needed IMHO as it is possible to change the special location with the "Change" option) with an "Show departures" one (copied from Leg-Station-Popup-Menu).

Also migrates `SpecialLocationPopupMenu` from Java to Kotlin.


app/src/main/res/menu/special_location_actions.xml
```diff
@@ -12,9 +12,9 @@
 		app:showAsAction="always"/>
 
 	<item
-		android:id="@+id/action_set_locations"
-		android:icon="@drawable/ic_mode_edit"
-		android:title="@string/action_set_locations"
+		android:id="@+id/action_show_departures"
+		android:icon="@drawable/ic_action_departures"
+		android:title="@string/find_departures"
```
grote: Was was this action doing before and why did you remove it?

ialokim: I think it was only putting the special location as "to", but did not query until you would change the "from" or "to" location. This makes sense for real favorite trips where you might want to change just one of the two "from" or "to" location for your query.

In the case of "home" and "work" however, where the "from" location is automatically filled with the current GPS location, it is easier to simply search for the station you want instead of using the drop-down menu option.

PR #663, opened 2020-02-19
Fix lint issues

ialokim:
Fixing several problems complained by Android Lint.


app/build.gradle
```diff
@@ -92,7 +92,9 @@ android {
 
 	lintOptions {
 		checkReleaseBuilds false
-		disable 'MissingTranslation', 'InvalidPackage', 'VectorDrawableCompat'
+		disable 'MissingTranslation', 'InvalidPackage', 'VectorDrawableCompat',
+				'TrustAllX509TrustManager', 'GradleDependency', 'IconDensities',
+				'PrivateResource', 'Typos', 'TypographyEllipsis'
```
grote: Maybe we should add a comment for the non-obvious ones and explain why we disable them? `GradleDependency` and `PrivateResource` for example are not clear to me.

app/src/androidTest/java/de/grobox/transportr/map/MapActivityTest.kt
```diff
@@ -85,6 +86,7 @@ class MapActivityTest : ScreengrabTest() {
     }
 
     @Test
+    @SuppressLint("IgnoreWithoutReason")
     @Ignore // TODO fix
```
grote: Maybe the better fix is to give a reason to the `@Ignore`?

app/src/main/AndroidManifest.xml
```diff
@@ -38,7 +38,7 @@
 		android:roundIcon="@mipmap/ic_launcher_round"
 		android:supportsRtl="true"
 		android:theme="@style/AppTheme.Light"
-		tools:ignore="AllowBackup,GoogleAppIndexingWarning">
+		tools:ignore="AllowBackup,GoogleAppIndexingWarning,UnusedAttribute">
```
grote: Which attribute is unused?

app/src/main/java/de/grobox/transportr/locations/LocationFragment.java
```diff
@@ -131,7 +130,7 @@ public View onCreateView(LayoutInflater inflater, @Nullable ViewGroup container,
 		linesLayout = v.findViewById(R.id.linesLayout);
 		linesLayout.setVisibility(GONE);
 		linesLayout.setAdapter(adapter);
-		linesLayout.setLayoutManager(new LinearLayoutManager(getContext(), HORIZONTAL, false));
+		linesLayout.setLayoutManager(new LinearLayoutManager(getContext(), RecyclerView.HORIZONTAL, false));
```
grote: What rule isn't allowing us to do a static import here?

app/src/main/java/de/grobox/transportr/trips/detail/TripDrawer.kt
```diff
@@ -165,7 +168,11 @@ internal class TripDrawer(context: Context) : MapDrawer(context) {
         val drawable: Drawable
         if (type == MarkerType.STOP) {
             drawable = ContextCompat.getDrawable(context, R.drawable.ic_marker_trip_stop) ?: throw RuntimeException()
-            drawable.mutate().setColorFilter(backgroundColor, PorterDuff.Mode.SRC_IN)
+            if (Build.VERSION.SDK_INT >= 29) {
+                drawable.mutate().colorFilter = BlendModeColorFilter(backgroundColor, BlendMode.SRC_IN)
+            } else {
+                drawable.mutate().setColorFilter(backgroundColor, PorterDuff.Mode.SRC_IN)
+            }
```
grote: We have these a couple of times. I wonder if there's no better way to do these. Maybe `DrawableCompat#setTint()`

app/build.gradle
ialokim: `GradleDependency` as we don't want the CI to fail just because of newer available versions of some gradle dependencies. `PrivateResource` is producing a lot of false positives e.g. complaining about R.string.app_name being a private resource of Mapbox. Also it seems some private resources from the google material design library are currently used but I didn't manage to copy them all.
`Typos` and `TypographyEllipsis` may be removed after the next translations fetch as I've corrected these strings in Transifex.

So yes, comments seem to be a good idea here.

app/src/androidTest/java/de/grobox/transportr/map/MapActivityTest.kt
ialokim: The problem is that I didn't know the reason, do you remember what exactly didn't work in this test?

app/src/main/java/de/grobox/transportr/locations/LocationFragment.java
ialokim: I guess the problem was that we included `HORIZONTAL` from a deprecated namespace in this place.

app/src/main/java/de/grobox/transportr/trips/detail/TripDrawer.kt
ialokim: Nice, I didn't find this one while being sure that it had to exist. Thanks, I'm going to change it.

app/src/main/AndroidManifest.xml
ialokim: `android:networkSecurityConfig` for older Android versions, but I guess that it is required for newer ones so we can't remove it here.

app/build.gradle
grote: cool and maybe add a TODO for those that we plan on getting rid of, but can't just yet.

`PrivateResource` sounds like something we should solve eventually.

app/src/androidTest/java/de/grobox/transportr/map/MapActivityTest.kt
grote: No, but you can just say currently broken ;)

app/src/main/AndroidManifest.xml
grote: strange, but if it is needed for newer versions, lint shouldn't complain about it!?

app/src/main/java/de/grobox/transportr/locations/LocationFragment.java
grote: So we can static import it from the new one again?

ialokim: I guess so. Is it better practice to do so?

app/src/main/AndroidManifest.xml
ialokim: Another option for lint to ignore this is adding `tools:targetApi="n"`. Apparently this really only affects lint. Do you want me to change it accordingly?

app/src/main/java/de/grobox/transportr/trips/detail/TripDrawer.kt
ialokim: Actually it turns out not being that easy, I've found several complaints about DrawableCompat not working as expected. For now, I would consider sticking to how it was before as this is at least guaranteed to work on all Android versions.

app/src/main/java/de/grobox/transportr/locations/LocationFragment.java
ialokim: Changed.

app/src/androidTest/java/de/grobox/transportr/map/MapActivityTest.kt
ialokim: Done.

app/build.gradle
ialokim: I've added the respective comments.

app/src/main/AndroidManifest.xml
ialokim: Changed to `tools:targetApi="n"`.

app/src/main/java/de/grobox/transportr/trips/detail/TripDrawer.kt
ialokim: Reverted to how it was before.

app/src/main/java/de/grobox/transportr/locations/LocationFragment.java
grote: It is a matter of taste. In times of modern IDEs that show you where things come from, I prefer static imports as they make the code smaller and easier to read.

app/src/main/java/de/grobox/transportr/trips/detail/TripDrawer.kt
grote: What a shame :(

Then maybe you can factor this if/else into a UiUtils method, so this drawable tinting is happening in one place only we can later change more easily?

ialokim: I would leave it as is for now, the IDE still complains about deprecated `setColorFilter` so it will be quite easy to find the (only four) places where it should be adapted later on.

Sincerely I don't want to change it for now as I fear to break some old Android versions without testing on various API versions. I guess this could be done in another PR in the future.

PR #664, opened 2020-02-20
Directions UI

### LINKED ISSUE #600: Direction search usability
Please reduce the clicks needed to adjust the search parameters in the direction search.

A bit of back story... I have been using this app since its first release on F-Droid when it was called Liberario. I love the app and a big thanks to grote and all the contributors! :blush:

My first reaction to the new GUI of version 2 was that I did not like it and I found #462 very quickly. However, I am aware that most GUI changes do not resonate well in the beginning, so I refrained from commenting and kept using the app.

In the last few weeks I had to travel more often. Although I knew the city, I did not know the public transport lines beyond the central area and hence, used the app extensively. The GUI has not grown on me. [ort163's comment about its usability](https://github.com/grote/Transportr/issues/462#issuecomment-407024175) is spot on.

I finally decided to create a new ticket, because of this comment in the issue mentioned above (and got convinced to do it now, because it will get a nice number :wink:):

> This are all for the directions search mask and should be a dedicated ticket I think.

Bringing back the GUI of version 1 would be an improvement because of the following reasons:

* Setting the time and arr/dep should be considered as a basic functionality and should be very visible.
  * For example, if you go to a location, you maybe need to set a specific arrival time.
  * If you go home on the other hand, you want to depart now.
  * So you may have to change those parameters for every new search.
* Better visibility should reduce the number of clicks needed and hence, improve the overall user experience.
* The display of the parameters is neither confusing or distracting, nor using up too much screen space.

Of course, we can only speculate about the usefulness for the majority of users. Unless there is a poll. However, the reactions to #462 indicate that there is significant interest. I propose that the reactions to this post should be considered as a primary source when evaluating the usefulness.

* [x] I am aware that Transportr is mostly developed by one person in their unpaid spare time.
* [ ] I can help myself to get this feature implemented or know someone who wants to do it.

I hope I could address all the topics in the feature request template.

ialokim:
Rearrange the directions form according to the discussion in #600.

directions_ui

Also tested on smaller screens where the timepicker popup is scrollable.

Fixes #600


app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
```diff
@@ -0,0 +1,274 @@
...
+        TooltipCompat.setTooltipText(productsIcon, getString(R.string.action_choose_products))
+        TooltipCompat.setTooltipText(swapIcon, getString(R.string.action_switch_locations))
+        TooltipCompat.setTooltipText(viaIcon, getString(R.string.action_navigation_expand))
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        super.onSaveInstanceState(outState)
+        outState.putSerializable(DATE, viewModel.calendar.value)
+        outState.putBoolean(EXPANDED, viewModel.isExpanded.value ?: false)
+        outState.putBoolean(DEPARTURE, viewModel.isDeparture.value ?: true)
```
grote: As a future optimization we can look into https://developer.android.com/topic/libraries/architecture/viewmodel-savedstate

app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
```diff
@@ -0,0 +1,274 @@
...
+        super.onViewStateRestored(savedInstanceState)
+        if (savedInstanceState != null && viewModel.trips.value == null) {
+            viewModel.setIsExpanded(savedInstanceState.getBoolean(EXPANDED, false))
+            viewModel.setIsDeparture(savedInstanceState.getBoolean(DEPARTURE, true))
+            viewModel.setFromLocation(fromLocation.getLocation())
+            viewModel.setViaLocation(viaLocation.getLocation())
+            viewModel.setToLocation(toLocation.getLocation())
+            viewModel.onTimeAndDateSet(savedInstanceState.getSerializable(DATE) as Calendar)
+        }
+        (activity!!.supportFragmentManager.findFragmentByTag(TimeDateFragment.TAG) as? TimeDateFragment)?.setTimeDateListener(viewModel)
```
grote: Not necessary now, but eventually we should get rid of listeners like this, they are flaky. Using the same ViewModel and LiveData is better.

app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
```diff
@@ -38,19 +39,22 @@ import java.util.Calendar.*
 class TimeDateFragment : DialogFragment(), OnDateSetListener, OnTimeChangedListener {
 
     private var listener: TimeDateListener? = null
+    private var departure: Boolean? = null
```
grote: If this is always set in `onCreate()` it could be `lateinit` and `Boolean` (non-null).

grote: Ah it being `null` means don't show departure stuff. Maybe add a quick documentation for that?

app/src/main/res/layout/fragment_directions_form.xml
```diff
@@ -142,4 +117,105 @@
...
+		android:id="@+id/timeIcon"
+		android:layout_width="@dimen/buttonSize"
+		android:layout_height="@dimen/buttonSize"
+		android:layout_marginTop="8dp"
+		android:background="?selectableItemBackgroundBorderless"
+		android:clickable="false"
+		android:focusable="false"
+		app:layout_constraintStart_toStartOf="parent"
+		app:layout_constraintTop_toBottomOf="@+id/toCard"
+		app:srcCompat="@drawable/ic_time"/>
```
grote: Below we use `android:src` instead. I remember this being tricky and can cause crashes on Android 4. Better check what we do elsewhere.

app/src/main/res/values-ca/strings.xml
```diff
@@ -52,8 +52,7 @@ i sempre saber on ets per no perdre on baixar de l\'autobús. ]]></string>
   <string name="action_fav_trip">Marcar trajecte com a favorit</string>
   <string name="action_unfav_trip">Desmarcar trajecte com a favorit</string>
   <string name="action_switch_locations">Intercanvia Des de i Cap a</string>
-  <string name="action_time_type">L\'hora és</string>
-  <string name="trip_dep">de sortida</string>
+	<string name="trip_dep">de sortida</string>
```
grote: Why do these use a different indentation? did you run tx pull --mode=developer?

artwork/ic_trip_arrival.svg
```diff
@@ -0,0 +1,58 @@
...
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="24"
+   height="24"
+   viewBox="0 0 24 24"
+   version="1.1"
+   id="svg6"
+   sodipodi:docname="ic_trip_arrival.svg"
```
grote: Are these SVGs still needed?

app/src/main/res/layout/fragment_directions_form.xml
```diff
@@ -142,4 +117,105 @@
...
+
+	<ImageView
+		android:id="@+id/productsMarked"
+		android:layout_width="wrap_content"
+		android:layout_height="wrap_content"
+		android:src="@drawable/ic_action_dot"
+		android:layout_marginLeft="16dp"
+		android:layout_marginStart="16dp"
+		android:layout_marginBottom="16dp"
+		android:tint="@color/md_grey_300"
```
grote: Does this work in the dark theme? Usually one uses a theme attribute like `?attr/colorControlNormal`, but one could also debate whether a red dot catches more attention.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
ialokim: Sounds interesting indeed.

app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
ialokim: > Ah it being `null` means don't show departure stuff. Maybe add a quick documentation for that?

Exactly, a little comment would be good here.

app/src/main/res/values-ca/strings.xml
ialokim: I noticed that too, it happened when deleting some now unused string resources via the Translation Editor in IntelliJ (s.th. the strings are automatically removed from all languages and Lint wouldn't complain about them). After this being merged to master, Transifex should pick the changes up and upon next `tx pull` the indentation should be consistent again.

artwork/ic_trip_arrival.svg
ialokim: No, good catch. I've only deleted them from `res`.

app/src/main/res/layout/fragment_directions_form.xml
ialokim: Well red wouldn't work that well I guess because the background is red already (in both dark and light theme). To be honest, I didn't try this part in the dark theme, I'll have a look at it.

app/src/main/res/layout/fragment_directions_form.xml
ialokim: Good catch. I'll check it out.

app/src/main/res/layout/fragment_directions_form.xml
ialokim: It shouldn't change anything in this case as the background is still red using the dark theme.

app/src/main/res/layout/fragment_directions_form.xml
ialokim: [It seems](https://android-developers.googleblog.com/2016/02/android-support-library-232.html) `app:srcCompat` should be the way to go, but also the following line should be added to `build.gradle`: `vectorDrawables.useSupportLibrary = true`. This should be done and tested in another PR though, I guess?

For now, it seems we are using both `android:src` and `app:srcCompat` without a clear distinction.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
ialokim: Opened #666

app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
ialokim: Moved to #667.

app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
ialokim: Added in updated commit.

artwork/ic_trip_arrival.svg
ialokim: Removed in updated commit.

app/src/main/res/layout/fragment_directions_form.xml
grote: Yeah that is the officially recommended way, but back then there were issues with it. I'd just do what is done elsewhere in Transportr for now and do any srcCompat change in a dedicated PR that needs testing on various phones especially those with minSdkVersion.

grote: I think that one is the last blocker for merging.

ialokim: We are currently using 23 times `app:srcCompat` and 9 times `android:src` throughout the app. So not sure what to pick here then.

grote: OMG! OK so let's just leave it like this and pray!

PR #675, opened 2020-03-17
Fix lifecycle dependencies

### LINKED ISSUE #662: Return to trip results freezes app
**Describe the bug**
At least on emulators, returning from a trip details view to the trip results list causes the app to freeze on current master. 

**To Reproduce**
1. Search for a connection
2. Click on one displayed result
3. Try to go back using the arrow in the top-left corner or using the phone's back button
4. See app freezing

**Expected behavior**
It should not freeze.

 - **Transportr Version:** current master
 - **Device:** emulators
 - **Android Version:** any version

This might be due to some library upgrade, perhaps also connected to Mapbox rendering problems on emulators?

ialokim:
As mentioned [here](https://developer.android.com/jetpack/androidx/releases/lifecycle#2.2.0), `lifecycle-extensions` is deprecated by version 2.2.0 and should be replaced by the needed Lifecycle artifacts (in this case just `lifecycle-livedata`).

Fixes #662 (somehow on the way :smile: ).


grote:
Thanks!

Issue #668, opened 2020-03-01
Update Mapbox SDK to latest version


ialokim:
Currently Transportr [uses Mapbox SDK 5.5.0](https://github.com/grote/Transportr/blob/master/app/build.gradle#L134), latest version is [8.6.2](https://www.mapbox.com/install/android/).

Are there any reasons why we didn't update the SDK for some time?

grote:
Updating even minor versions of Mapbox has been a pain and huge time-sink. The APIs changed all the time and required various code changes. Not sure if this has changed by now. Feel free to update to the latest version, but please make sure to test things thoroughly afterwards, especially GPS related things like GPS disabled, enabling, stopping activitiy, starting it again, following a trip with GPS and see if this works reliably.

ialokim:
It seems, the newest version of Mapbox SDK is anyhow [not fully free software anymore](https://forum.f-droid.org/t/updated-distribution-license-for-mapbox-mobile-maps-sdk/10201), so we might stick to the current used version for now. For the future, we might consider switching the map library, two alternatives to consider might be [Tangram](https://github.com/tangrams/tangram-es) (formally part of Mapzen, now a Linux Foundation project) or [Carto Mobile SDK](https://github.com/CartoDB/mobile-sdk).

ialokim:
The currently used Mapbox SDK version also has problems in combination with latest `androidx.fragment` package: Something seems to have changed in the way `onDestroyView()` is called and therefore also `mapView.onDestroy()`. This results in an app freeze with the following ANR stack trace:

<details>
    <summary>Click to expand</summary>

```
DALVIK THREADS (26):
"main" prio=5 tid=1 Native
  | group="main" sCount=1 dsCount=0 obj=0x73888000 self=0xb3c25800
  | sysTid=16563 nice=0 cgrp=default sched=0/0 handle=0xb7706ea0
  | state=S schedstat=( 0 0 0 ) utm=98 stm=47 core=1 HZ=100
  | stack=0xbf475000-0xbf477000 stackSize=8MB
  | held mutexes=
  kernel: futex_wait_queue_me+0xcd/0x113
  kernel: futex_wait+0xc5/0x197
  kernel: do_futex+0x9b/0x71e
  kernel: SyS_futex+0xaf/0xf6
  kernel: syscall_call+0x7/0xb
  native: #00 pc 000132d0  /system/lib/libc.so (syscall+32)
  native: #01 pc 00000c47  /system/bin/linker (???)
[...]

PR #677, opened 2020-03-23
Automatic theming and fix some preference issues

### LINKED ISSUE #578: Theme does not switch properly on changing
**Describe the bug**
When switching the theme (from light to dark or vice-versa), the settings activity is automatically closed, but the main (map) activity does not reflect the newly selected theme.

**To Reproduce**
1. Go to the app's settings
2. Switch the theme
3. Note that the settings view disappears, but the theme seems to be unchanged
4. On closing and opening Transportr again, the new theme is properly displayed

**Expected behavior**
The theme should switch right away.

 - **Transportr Version:** 2.0.5
 - **Device:** Wileyfox Swift
 - **Android Version:** 7.1.2

### LINKED ISSUE #673: Automatic theme matching system theme
Several apps are able to select light or dark theme automatically matching the current Android system-wide theming. Would be nice if Transportr could do the same.

### LINKED ISSUE #676: Changing language in-app does not change title in toolbar
After changing the language to use inside the app's settings, the toolbar titles are not adjusted to the new language but still use the system's default.

Screenshot_1584728283

ialokim:
This PR changes the app's theme to be based on a `DayNight` theme to support automatic dark theming for new Android versions as [described in the Android docs](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#themes_and_styles). Changing the theme now works without restarting the activities (except the map theme which is currently not updated immediately in the `MapActivity` after changing the app's theme in settings (therefore fixes #578). However, upon restart of the app, the map theme is correctly set. I would prefer to fix this particular issue in another PR (after) updating to the newest Mapbox SDK version (see #668).

Also language switcher now works as expected (by restarting the app's main activity upon change) and adapting the actionbar title too. Therefore, fixes #676.

Fixes #673.


app/build.gradle
```diff
@@ -107,10 +107,10 @@ android {
 dependencies {
 	implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
 
-	implementation 'androidx.appcompat:appcompat:1.1.0'
-	implementation('androidx.preference:preference:1.1.0-alpha01') {
-		because 'fixes bug with preference categories'
+	implementation('androidx.appcompat:appcompat:1.2.0-alpha03') {
+		because 'fixes bug where stopped activities would not resume'
```
grote: :man_facepalming:

grote: Risky taking an alpha version here, but I guess you don't intent to make a release soon?

app/src/main/res/values-night/themes.xml
```diff
@@ -0,0 +1,34 @@
+<resources>
+
+    <!-- App Theme -->
+
+    <style name="AppDayNightTheme" parent="AppBaseTheme">
+		<item name="colorControlNormal">@color/drawableTintDark</item>
```
grote: intentation wrong

grote:
Thanks for the PR! I guess you tested this thoroughly and made sure there's no new bugs?

app/build.gradle
ialokim: True, but otherwise DayNight theme switching would [not have worked on API level 23 and below](https://developer.android.com/jetpack/androidx/releases/appcompat#1.2.0-alpha03). The changes since 1.1.0 seem to be only bugfixes. I would not want to make a release very soon, but perhaps before 1.2.0 is stable though. After testing it heavily on different API levels, we could perhaps release as is (similar to what we did with `androidx.preference` apparently)?

app/src/main/res/values-night/themes.xml
ialokim: Good catch, changed in `values-v29/themes.xml` too.

Issue #686, opened 2020-07-23
Popup theme not properly changed


ialokim:
On current master and (at least) API 22, popups are always displayed using dark theme no matter which app theme is picked:

![Screenshot_1595537448](https://user-images.githubusercontent.com/18088991/88337490-1a846100-cd37-11ea-8004-50b73378b089.png)

grote:
Maybe the dialog theme being used is either not properly declared as a theme attribute or declared when launching the dialog, or the dialog theme itself is not a proper DayNight theme.

PR #687, opened 2020-08-02
Fix theming issues

### LINKED ISSUE #685: MapStyle not changed directly after switching theme
As already mentioned in #677, changing the app's theme does not change the map style directly. It is currently only changed after an app restart. It might make sense to address this issue after #668.

ialokim:
Fixes #685 and #686 in two separate commits.


grote:
Thanks!

PR #690, opened 2020-08-05
Add template for releases

ialokim:
... as documentation for the release process.


grote:
Awesome, this is the perfect companion to the release script I am currently working on that handles all that comes after your check list.

PR #710, opened 2020-09-14
Close several prioritized but minor issues

ialokim:
See the commit messages for further information.


grote:
Thanks, looks good!

Would you mind fixing the lint gradle task before the next PR. Now it fails with OOM error.

PR #712, opened 2020-09-18
Increase JVM memory and fix lint errors

ialokim:
Not sure how to start Travis CI from a forked repository and as the Lint OOM exception doesn't happen on my machine, testing simply to increase the available JVM memory.


grote:
Is the `:app:lint` task completing on your machine?

ialokim:
Yes, it takes its time to complete (maybe 1-2 minutes), but it always does.

It's always this `Caused by: java.lang.OutOfMemoryError: GC overhead limit exceeded` error. Most probably this is related to the gradle and android-gradle-plugins version upgrade in 862182156fe19b5fc860dbd68388ff299f9b4791, but I still didn't find anything useful in the internet.

We might also consider simply omitting the lint task on build?

ialokim:
See also
- https://guides.codepath.com/android/Setting-up-Travis-CI#troubleshooting
- https://stackoverflow.com/questions/27164452/how-to-solve-java-lang-outofmemoryerror-gc-overhead-limit-exceeded-error-in-and
- https://github.com/gradle/kotlin-dsl-samples/pull/953
- https://github.com/broadinstitute/gatk/pull/1124/files

app/build.gradle
```diff
@@ -57,7 +57,7 @@ android {
 
 	dexOptions {
 		preDexLibraries !project.hasProperty('buildServer')
-		javaMaxHeapSize "1g"
+		javaMaxHeapSize "2g"
```
grote, line 60: Maybe try to remove those dexOptions completely. They might not be needed anymore. Also, the max heap size here might not affect the lint process. That's gradle options.

ialokim:
Increasing the JVM memory for the gradlew daemon using [`org.gradle.jvmargs`](https://docs.gradle.org/4.0/userguide/build_environment.html#sec:gradle_configuration_properties) did the trick.

app/build.gradle
```diff
@@ -92,6 +92,7 @@ android {
 
 	lintOptions {
 		checkReleaseBuilds false
+		abortOnError false
```
grote, line 95: Is this still needed? I think better to disable the lint checks that fail for now until we can resolve them.

ialokim: Yeah, this is probably better. I've addressed the lint errors in the force-pushed version of the branch.

grote:
Nice, looks good!

PR #713, opened 2020-09-23
Improved geo intents

### LINKED ISSUE #698: Hide Transportr from "Show on External Map" (otherwise error appears)
**Describe the bug**
When chosen Transportr to show as external map, `Couldn't open external location` error appears. Logically, as it's not an **external** map. Could Transportr be hidden from that menu?

**To Reproduce**
1. Go to travel
2. Tap on a line/way/plan
3. Tap on the three points of a place
4. Tap on `Show on External Map`
5. Choose Transport
6. `Couldn't open external location` error appears

**Expected behavior**
Transportr should **not** appear.

**Screenshots**
transportr-external-map

 - **Transportr Version:** 2.0.1
 - **Device:** Xiaomi Redmi Not 4 X
 - **Android Version:** /e/OS Nougat (7.1.2)

ialokim:
- recognize `geo:0,0?q=lat,lon` intents as per [Android docs](https://developer.android.com/guide/components/intents-common\#ViewMap) and as used by Transportr itself
- exclude Transportr from "show on external map"

Fixes #698

### DIFF
--- a/app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
+++ b/app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
@@ -97,11 +97,11 @@ object IntentUtils {
 
     @JvmStatic
     fun getWrapLocation(geoUri: String): WrapLocation? {
-        val pattern = Pattern.compile("^geo:(-?\\d{1,3}(\\.\\d{1,8})?),(-?\\d{1,3}(\\.\\d{1,8})?).*")
+        val pattern = Pattern.compile("^geo:(0,0\\?q=)?(-?\\d{1,3}(\\.\\d{1,8})?),(-?\\d{1,3}(\\.\\d{1,8})?).*")
         val matcher = pattern.matcher(geoUri)
         if (matcher.matches()) {
-            val lat: Double = matcher.group(1)!!.toDouble()
-            val lon: Double = matcher.group(3)!!.toDouble()
+            val lat: Double = matcher.group(2)!!.toDouble()
+            val lon: Double = matcher.group(4)!!.toDouble()
             return if (lat == 0.0 && lon == 0.0) null else WrapLocation(LatLng(lat, lon))
         }
         return null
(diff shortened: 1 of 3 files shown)


app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
```diff
@@ -86,22 +96,33 @@ object IntentUtils {
...
-        intent.flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
-        intent.data = geo
-        if (intent.resolveActivity(context.packageManager) != null) {
-            Log.d(context.javaClass.simpleName, "Starting geo intent: " + geo.toString())
-            context.startActivity(intent)
+        val intent = Intent(Intent.ACTION_VIEW).apply {
+            flags = Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_SINGLE_TOP
+            data = geo
+        }
+        val intents = context.packageManager.queryIntentActivities(intent, 0).mapNotNull {
```
grote, line 103: bad news: https://cketti.de/2020/09/03/avoid-intent-resolveactivity/

but can stay like this in this PR as it was like this before.

app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
```diff
@@ -86,22 +96,33 @@ object IntentUtils {
...
+            val packageName = it.activityInfo.packageName
+            if (context.packageName != packageName)
+                Intent(intent).apply { `package` = packageName }
+            else null
```
grote, line 107: Do we need to set the intent's package name here?

grote (approved):
Thanks for the PR!

app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
ialokim: Yeah, we might stumble upon this again when targeting Android 11.

app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
ialokim: Yes, that is how it was done in https://stackoverflow.com/questions/38345994/how-to-exclude-a-specific-application-from-action-send-intent.

Basically, we have multiple intents filtering for one specific `packageName` each to be passed in `Intent.EXTRA_INITIAL_INTENTS`.

For API >= 24 it would have been simpler by just using `Intent.EXTRA_EXCLUDE_COMPONENTS`.

PR #716, opened 2020-09-29
Refactor several DateUtils, update (relative) times in trip details and directions form

### LINKED ISSUE #700: "in x minutes" does not update
**Describe the bug**
~When searching for connections and~ In the trip detail view, you can see a "in x minutes"/"now"/"x minutes ago" text. After waiting for e.g. one minute (turning the display off or leaving it on), the displayed text should adapt to the new current time. This does not happen and is especially confusing if you search for a connection, go to the first platform, turn your display off, get on the train (e.g.) and look for your connection after some minutes again, it will still report the time before getting on the train.

**To Reproduce**
1. Search for a connection
2. Select a specific connection within the next minutes ~(optional)~
3. Stay on the same screen for some time (or turn your display off and on again after some time)
4. Notice that the time description does not change over time until reloading.

**Expected behavior**
The string should update according to the current time.

 - **Transportr Version:** 2.1.0
 - **Device:** Wileyfox Swift
 - **Android Version:** 9

ialokim:
See individual commits for more information. They might be squashed before merging, but I preferred to keep them separate for easier review.

Fixes #700.


grote:
Wow, thanks a lot! Do you have screenshots for the UI visible changes?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
```diff
@@ -0,0 +1,209 @@
...
+    val findGpsLocation = MutableLiveData<FavLocationType?>()
+    val timeUpdate = LiveTrigger()
+    private val _now = MutableLiveData<Boolean>()
+    private val _calendar = MutableLiveData<Calendar?>()
+    private val _products = MutableLiveData<EnumSet<Product>>()
+    private val _isDeparture = MutableLiveData<Boolean>()
+    private val _isExpanded = MutableLiveData<Boolean>()
+    val showTrips = SingleLiveEvent<Void>()
+    @JvmField
+    val topSwipeEnabled = MutableLiveData<Boolean>()
```
grote, line 68: why does this need a JvmField annotation?

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
```diff
@@ -0,0 +1,209 @@
...
+    val queryMoreError: LiveData<String>
+        get() = _tripsRepository.queryMoreError
+
+    init {
+        _now.value = true
+        _calendar.value = Calendar.getInstance()
+        _products.value = EnumSet.allOf(Product::class.java)
+        _isDeparture.value = true
+        _isExpanded.value = false
+        topSwipeEnabled.value = false
```
grote, line 199: You can set these initial values in the live data constructor instead.

app/src/main/java/de/grobox/transportr/trips/search/TripsFragment.kt
```diff
@@ -0,0 +1,175 @@
...
+import de.grobox.transportr.ui.LceAnimator
+import de.grobox.transportr.utils.Linkify
+import de.grobox.transportr.utils.TransportrUtils.getDragDistance
+import de.schildbach.pte.dto.Trip
+import kotlinx.android.synthetic.main.fragment_trips.*
+import java.util.regex.Pattern
+import javax.annotation.ParametersAreNonnullByDefault
+import javax.inject.Inject
+
+@ParametersAreNonnullByDefault
```
grote, line 49: I think this annotation isn't needed in Kotlin.

app/src/test/java/de/grobox/transportr/utils/DateUtilsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.Assert
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.Mockito.*
+import org.mockito.MockitoAnnotations
+import java.util.*
+
+
+class DateUtilsTest {
```
grote, line 35: Nice!

app/src/test/java/de/grobox/transportr/utils/DateUtilsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+    @Mock
+    lateinit var resources: Resources
+
+    private fun minuteToMillis(minutes: Float): Long {
+        return (minutes * android.text.format.DateUtils.MINUTE_IN_MILLIS).toLong()
+    }
+
+    @Test
+    fun millisToMinutes() {
+        Assert.assertEquals(0, DateUtils.millisToMinutes(minuteToMillis(0f)))
```
grote, line 48: You could import the methods directly, so removing `Assert.` and `DateUtils.`. Makes the code a bit smaller and easier to read, I'd say.

app/src/test/java/de/grobox/transportr/utils/IntentUtilsTest.kt
```diff
@@ -25,7 +25,7 @@ import org.junit.Assert
 import org.junit.Test
 
 
-class MapViewModelTest {
+class IntentUtilsTest {
```
grote, line 28: Thanks!

grote (changes requested):
Awesome, thanks for the PR!

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
ialokim: Good question, I guess the IntelliJ converter added this. Will recheck.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
ialokim: True, might be cleaner.

app/src/main/java/de/grobox/transportr/trips/search/TripsFragment.kt
ialokim: Again, going to check. I'm not very familiar with these annotations.

app/src/test/java/de/grobox/transportr/utils/DateUtilsTest.kt
ialokim: Sounds like a good idea :+1:

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
ialokim: Removed.

app/src/main/java/de/grobox/transportr/trips/search/TripsFragment.kt
ialokim: Removed too.

ialokim:
I've just incorporated your comments.

Here are screenshots before the changes:
old

and after:
new

Note the slightly more natural spacing between "Departure" and "Now" or the date/time. Also note that not everything within a ten minutes time-frame is considered as "Now" anymore, instead it shows relative times up to 30 minutes in past or future.

grote:
> Note the slightly more natural spacing between "Departure" and "Now" or the date/time.

Now that they are so close, I wonder if `Now` should be `now` instead. Or `Departure:` with a colon?

> Also note that not everything within a ten minutes time-frame is considered as "Now" anymore,

Was that the case before? 10 minutes sounds a lot for "now".

> instead it shows relative times up to 30 minutes in past or future.

Do you mean seconds or minutes?

ialokim:
> Now that they are so close, I wonder if `Now` should be `now` instead. Or `Departure:` with a colon?

Good point, I had been thinking about the same thing but forgot it at the end. I would prefer `now` instead of the colon as `in x minutes`/`x minutes ago` is also written in lower-case letters.

> Was that the case before? 10 minutes sounds a lot for "now".

Indeed, see here: https://github.com/grote/Transportr/blob/707bea65f8de38caf29089fdc1508637261c0087/app/src/main/java/de/grobox/transportr/utils/DateUtils.java#L94

> Do you mean seconds or minutes?

I mean minutes. If you search for a trip within the next (or previous) 30 minutes, a relative time information (e.g. `Departure in 25 minutes`) will be shown, before or after that timeframe it falls back to the normal time (e.g. `Departure 10:05 AM`).

grote:
>  I would prefer now instead of the colon

Fine with me



> > Was that the case before? 10 minutes sounds a lot for "now".
>>
> Indeed

😯

> it shows relative times up to 30 minutes in past or future.

I would have shown the relative times for longer, but 30min is also fine.

Issue #706, opened 2020-09-09
Connection search forgets input of "From" field


AltNico:
**To Reproduce**

* Open Transportr
* Enter some destination into "Search here", e.g. "Hauptbahnhof, Frankfurt (Main)" in the RMV network
* Click on red circle at bottom right
* Enter some departure into "From", e.g. "Hauptbahnhof, Mainz"
* Click on some connection to get to the details page
* Use the Android back button to go to the list

**Expected behavior**

* "From" field still contains "Hauptbahnhof, Mainz"

**Actual behavior**

* "From" field is empty
* Loading more connections into both time directions still work, though, even after rotating
* However, upon changing filters like departure time or transportation modes, it doesn't initiate a new search. Instead, it's waiting for users to enter something into "From"

**Versions**

 - **Transportr Version:** 2.1.0
 - **Device:** LG-D802
 - **Android Version:** 10
 - **LineageOS Version:** 17.1

**Logcat output**

Here's the output of _adb logcat_:

```bash
$ adb logcat | grep -F "`adb shell ps | grep de.grobox.  | tr -s [:space:] ' ' | cut -d' ' -f2`"
09-09 18:06:28.261  1435  1435 D Zygote  : Forked child process 1827
09-09 18:06:28.266  1461  1721 I ActivityManager: Start proc 1827:de.grobox.liberario/u0a133 for pre-top-activity {de.grobox.liberario/de.grobox.transportr.map.MapActivity}
09-09 18:06:28.357  1827  1827 V AndroidLocationEngine: Initializing.
09-09 18:06:28.358  1827  1827 V MapboxTelemetry: Initializing telemetry.
09-09 18:06:28.364  1827  1827 D NetworkSecurityConfig: Using Network Security Config from resource network_security_config debugBuild: false
09-09 18:06:28.386  1827  1827 V MapboxTelemetry: Disabling telemetry.
09-09 18:06:28.386  1827  1827 D MapboxTelemetry: Shutting down telemetry service.
09-09 18:06:28.393  1827  1827 I robox.liberari: Thread[1,tid=1827,Native,Thread*=0xb3100e00,peer=0x723211f0,"main"] recursive attempt to load library "/data/app/de.grobox.liberario-KBS5ZhV_anmfK-co2N5C-Q==/lib/arm/libmapbox-gl.so"
09-09 18:06:28.397  1827  1827 I chatty  : uid=10133(de.grobox.liberario) identical 1 line
[...]

PR #722, opened 2020-10-30
Fix double swap locations bug

ialokim:
Fixes the bug described in https://github.com/grote/Transportr/issues/706#issuecomment-719488100.


grote (approved):
Thanks for the debugging and the fix!

Issue #265, opened 2016-11-29
Feature Request: Support for Tor network (via Orbot)


github-user-1:
Something nice and straightforward to implement that I would like to see is (optional) support for Tor (via Orbot). It adds some additional level of anonymity and should be (I think) easy to implement - it requires to route through the proxy provided by Orbot (additionally the app could check if Orbot is running already or start it).

grote:
It might not be as easy as you might think since all HTTP connections are made through the public-transport-enabler library and not directly by Transportr. However, this is definitely something worth looking into.

grote:
With https://github.com/schildbach/public-transport-enabler/commit/86b2d5a2f3c8e0a03f8d59cb3dad1b5ee04679e8 this might now be possible.

full-duplex:
As far as I can tell there are two ways network connections are used:
- The public-transport-enabler providers: With the commit linked above that implements the `setProxy()` function for the `AbstractNetworkProvider` this is easy. (I don't think there are any providers that don't extend the `AbstractNetworkProvider`, but that needs to be checked)
- The map: The new beta uses the Mapbox Android SDK, which has no support for proxies (see [here](https://github.com/mapbox/mapbox-gl-native/blob/d169ea55c1cfa85cd8bf19f94c5f023569f71810/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/http/HTTPRequest.java). I don't know if this is the only place they do networking), so this will leak your traffic or will have to be disabled. (The old map provider, osmdroid, has a `setHttpProxy()` configuration option (and while we are on the subject of staying anonymous, doesn't do any telemetry))

Are there any other places where networking takes place that I forgot?

grote:
> I don't think there are any providers that don't extend the AbstractNetworkProvider

There aren't any.

> The new beta uses the Mapbox Android SDK, which has no support for proxies

That is too bad. Maybe someone can make a PR or request proxy support.

> and while we are on the subject of staying anonymous, doesn't do any telemetry

Yes, in this regard, the library is better. Unfortunately, it isn't in others. The mapbox telemetry *should* be disabled however.

> Are there any other places where networking takes place that I forgot?

I can't think of any right now.

grote:
> I can't think of any right now.

There are the reverse geocoder requests in Transportr 2. There's two different geocoders. When Tor is used, the Google one should be blocked and only the one where we control the HTTP request should be used.

blacklight447:
[...]

PR #726, opened 2020-11-06
Implement proxy preferences

micaplush:
This commit adds proxy preferences to the settings screen that are
then used in the ReverseGeocoder, MapBox and public-transport-enabler
(on NetworkProviders that support it, which should apply to every
provider).

One can verify that the app does not establish a direct internet
connection using a tool like nethogs (in Termux).

### DIFF
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
@@ -29,6 +29,8 @@ import androidx.core.app.ActivityOptionsCompat
 import androidx.lifecycle.Observer
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
+import com.google.common.net.HostSpecifier
+import com.google.common.net.InternetDomainName
 import de.grobox.transportr.R
 import de.grobox.transportr.TransportrApplication
 import de.grobox.transportr.map.MapActivity
@@ -41,6 +43,7 @@ import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_HOST
 import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PORT
 import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PROTOCOL
 import de.grobox.transportr.settings.SettingsManager.Companion.THEME
+import de.grobox.transportr.ui.ValidatedEditTextPreference
 import de.grobox.transportr.utils.updateGlobalHttpProxy
 import javax.inject.Inject
 
@@ -107,6 +110,25 @@ class SettingsFragment : PreferenceFragmentCompat() {
                     true
                 }
             }
+        }
+
+        findPreference<ValidatedEditTextPreference>(PROXY_HOST)!!.validate = { value ->
+            /*
+            FIXME: These validation functions are marked unstable, maybe find a replacement
+
+             Valid should be everything accepted by InetAddress.getByName(...),
+             that is: IPv4 addresses, IPv6 addresses, hostnames and FQDNs.
+
+             Additionally, no network requests should be performed as part of
+             the validation, so simply using InetAddress.getByName(...) with a
+             try/catch is out of the question since it might try to resolve
+             domain names.
+             */
+            InternetDomainName.isValid(value) || HostSpecifier.isValid(value)
+        }
+
+        findPreference<ValidatedEditTextPreference>(PROXY_PORT)!!.validate = { value ->
+            value.toIntOrNull() in 1..65534
         }
     }
 
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
@@ -23,7 +23,9 @@ import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_CLEAR_TASK
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.os.Bundle
+import android.util.Log
 import android.util.Patterns
+import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.app.AppCompatDelegate.*
 import androidx.core.app.ActivityCompat
 import androidx.core.app.ActivityOptionsCompat
@@ -43,7 +45,7 @@ import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PORT
 import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PROTOCOL
 import de.grobox.transportr.settings.SettingsManager.Companion.THEME
 import de.grobox.transportr.ui.ValidatedEditTextPreference
-import de.grobox.transportr.utils.updateGlobalHttpProxy
+import de.grobox.transportr.utils.TransportrUtils
 import javax.inject.Inject
 
 class SettingsFragment : PreferenceFragmentCompat() {
@@ -104,8 +106,22 @@ class SettingsFragment : PreferenceFragmentCompat() {
 
         arrayOf(PROXY_ENABLE, PROXY_HOST, PROXY_PORT, PROXY_PROTOCOL).forEach { prefKey ->
             (findPreference(prefKey) as Preference?)?.let { pref ->
-                pref.setOnPreferenceChangeListener { _, _ ->
-                    updateGlobalHttpProxy(settingsManager.proxy, manager)
+                pref.setOnPreferenceChangeListener { _, newValue ->
+                    try {
+                        val newProxy = settingsManager.getProxy(mapOf(
+                            Pair(prefKey, newValue)
+                        ))
+                        TransportrUtils.checkInternetConnectionViaProxy(newProxy)
+                        TransportrUtils.updateGlobalHttpProxy(newProxy, manager)
+                    } catch (e: Exception) {
+                        Log.e(TAG, "Invalid proxy settings: " + e.message)
+                        AlertDialog.Builder(context!!)
+                            .setTitle(R.string.invalid_proxy_settings)
+                            .setMessage(e.message)
+                            .setPositiveButton(android.R.string.ok) { di, _ -> di.dismiss() }
+                            .create()
+                            .show()
+                    }
                     true
                 }
             }
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
@@ -36,7 +36,12 @@ import de.grobox.transportr.networks.PickTransportNetworkActivity
 import de.grobox.transportr.networks.TransportNetwork
 import de.grobox.transportr.networks.TransportNetworkManager
 import de.grobox.transportr.settings.SettingsManager.Companion.LANGUAGE
+import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_ENABLE
+import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_HOST
+import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PORT
+import de.grobox.transportr.settings.SettingsManager.Companion.PROXY_PROTOCOL
 import de.grobox.transportr.settings.SettingsManager.Companion.THEME
+import de.grobox.transportr.utils.updateGlobalHttpProxy
 import javax.inject.Inject
 
 class SettingsFragment : PreferenceFragmentCompat() {
@@ -44,6 +49,9 @@ class SettingsFragment : PreferenceFragmentCompat() {
     companion object {
         val TAG: String = SettingsFragment::class.java.simpleName
     }
+
+    @Inject
+    internal lateinit var settingsManager: SettingsManager
 
     @Inject
     internal lateinit var manager: TransportNetworkManager
@@ -89,6 +97,15 @@ class SettingsFragment : PreferenceFragmentCompat() {
             it.setOnPreferenceChangeListener { _, _ ->
                 reload()
                 true
+            }
+        }
+
+        arrayOf(PROXY_ENABLE, PROXY_HOST, PROXY_PORT, PROXY_PROTOCOL).forEach { prefKey ->
+            (findPreference(prefKey) as Preference?)?.let { pref ->
+                pref.setOnPreferenceChangeListener { _, _ ->
+                    updateGlobalHttpProxy(settingsManager.proxy, manager)
+                    true
+                }
             }
         }
     }
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
@@ -23,14 +23,13 @@ import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_CLEAR_TASK
 import android.content.Intent.FLAG_ACTIVITY_NEW_TASK
 import android.os.Bundle
+import android.util.Patterns
 import androidx.appcompat.app.AppCompatDelegate.*
 import androidx.core.app.ActivityCompat
 import androidx.core.app.ActivityOptionsCompat
 import androidx.lifecycle.Observer
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
-import com.google.common.net.HostSpecifier
-import com.google.common.net.InternetDomainName
 import de.grobox.transportr.R
 import de.grobox.transportr.TransportrApplication
 import de.grobox.transportr.map.MapActivity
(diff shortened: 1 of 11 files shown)


T0astBread:
I accept the CLA  ✔️

T0astBread:
More on the nethogs thing: If you have a rooted phone you can run nethogs as root in Termux to see when an app establishes a network connection. This is how using a debug build without a proxy looks:

Screenshot of nethogs showing "de.grobox.liberari..." as one of its entries

and this is with proxy preferences set to use Tor:

Screenshot of nethogs not showing anything that looks like the Transportr app but with increased bandwidth usage from Tor

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
```diff
@@ -92,6 +96,26 @@ class SettingsManager @Inject constructor(private val context: Context) {
...
+                val type = when (typeStr) {
+                    "SOCKS" -> Proxy.Type.SOCKS
+                    "HTTP" -> Proxy.Type.HTTP
+                    else -> throw IllegalStateException("Illegal proxy type: " + typeStr)
+                }
+                val host = settings.getString(PROXY_HOST, null)
+                val port = Integer.parseInt(settings.getString(PROXY_PORT, null)!!)
+                Proxy(type, InetSocketAddress(InetAddress.getByName(host), port))
+            } catch (e: Exception) {
+                e.printStackTrace()
```
grote, line 114:
```suggestion
                Log.e(TAG, "Error parsing proxy settings", e)
```

app/src/main/java/de/grobox/transportr/utils/TransportrUtils.kt
```diff
@@ -114,3 +119,16 @@ object TransportrUtils {
...
+    // MapBox
+    HttpRequestUtil.setOkHttpClient(
+        OkHttpClient.Builder()
+            .proxy(newProxy)
+            .build())
+    // public-transport-enabler
+    manager.transportNetwork.value?.let {
+        if (it.networkProvider is AbstractNetworkProvider)
+            (it.networkProvider as AbstractNetworkProvider).setProxy(newProxy)
+    }
```
grote, line 133: Should we maybe add the proxy to PTE where its provider gets created instead?

grote (changes requested):
Awesome! Thanks a lot for your contribution! :smiley: 

Looks like this would close https://github.com/grote/Transportr/issues/265

app/src/main/java/de/grobox/transportr/utils/TransportrUtils.kt
T0astBread: I've tried to do that at first but it seems like the providers are created in a file called `de.grobox.transportr.networks.TransportNetworks` as part of a private val called `networks`, so I'm not sure where to take the proxy configuration from in that case. (Unless I've got something mixed up?)

It would work though (I've done it with hard-coded values) and I agree, it would probably be cleaner.

grote: Maybe add the proxy settings as a lamda parameter here:

https://github.com/grote/Transportr/blob/7b1866e15ae3cb26a31bb69195bc80dcf51c30c1/app/src/main/java/de/grobox/transportr/networks/TransportNetwork.kt#L42

T0astBread: I've tried to implement something like this now but I'm having some problems with it. To not spam this thread, I've made the commit in a new branch: https://github.com/T0astBread/Transportr/commit/4bb422b207aac951530a372bce3ecaf58fcb1f49.

That commit crashes when I do a location search with the following stacktrace:
```
2020-11-17 13:11:12.395 9655-9655/de.grobox.liberario.debug E/AndroidRuntime: FATAL EXCEPTION: main
    Process: de.grobox.liberario.debug, PID: 9655
    kotlin.UninitializedPropertyAccessException: lateinit property settingsManager has not been initialized
        at de.grobox.transportr.locations.LocationView.getSettingsManager$app_debug(LocationView.kt:66)
        at de.grobox.transportr.locations.LocationView$startSuggestLocationsTaskDelayed$1.run(LocationView.kt:239)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6718)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:491)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```

I think there are a few places in the code where dependency injection won't run before a SettingsManager is needed for the proxy and I'm not really sure what to do about that.

T0astBread: Oh and I've also not made the proxy a parameter to that factory function directly because you'd have to copy+paste `setProxy` in the `TransportNetworks` file for every provider in that case.

grote: Injecting stuff in Views might be trickier than in Activities and Fragments. Thanks for trying. If the first solution also handles network changes, I'd be fine with me, but let's also see what @ialokim thinks.

T0astBread: I've just tried, it seems to work across network changes

ialokim: I think that would be fine with me too, as a network change triggers the MainActivity to be recreated also the proxy settings are applied.

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
```diff
@@ -92,6 +97,26 @@ class SettingsManager @Inject constructor(private val context: Context) {
...
+            return try {
+                val typeStr = settings.getString(PROXY_PROTOCOL, null)
+                val type = when (typeStr) {
+                    "SOCKS" -> Proxy.Type.SOCKS
+                    "HTTP" -> Proxy.Type.HTTP
+                    else -> throw IllegalStateException("Illegal proxy type: " + typeStr)
+                }
+                val host = settings.getString(PROXY_HOST, null)
+                val port = Integer.parseInt(settings.getString(PROXY_PORT, null)!!)
+                Proxy(type, InetSocketAddress(InetAddress.getByName(host), port))
```
ialokim, line 113: What kind of hosts are accepted here? E.g. do they need `http(s)://`, IP adresses and/or domains?

ialokim (changes requested):
Thanks a lot for this PR! I think it would be nice to have some sort of user-feedback right away to see if the proxy settings they entered do work correctly. Not sure though where to hook this in: After a change to some of the fields or before leaving the settings screen?

Would you mind posting a screenshot of how the settings look like with these additions (in active and inactive proxy modes)?

T0astBread:
This is what the proxy settings currently look like:

proxy-settings-on
proxy-settings-off

As for validation, I think it would make it more clear what's wrong if something happened right after a bad value is entered. Ideally the OK button would be disabled all together when the state of the text input is not acceptable but I'm not sure that's possible with an EditTextPreference. I'll look into how this works.

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
T0astBread: IP addresses and domains, the protocol and port are handled by other inputs. The default is 127.0.0.1.

ialokim:
> This is what the proxy settings currently look like

Looks good to me, but I would just prefer a switch over the checkbox for enabling the proxy as documented here: https://source.android.com/devices/tech/settings/settings-guidelines#checkbox

> As for validation, (...) I'll look into how this works.

Thanks for investigating!

T0astBread:
Implemented some validation now. When the user enters an invalid value for either the host or port the "OK" button on the edit dialog is disabled immediately. The implementation is a bit more complicated than I'd like it to be since Android's EditTextPreference doesn't support that, so I kinda just wrote my own replacement for it.

T0astBread:
It should also be noted that the validation doesn't match exactly the same cases as the proxy implementation can handle but, as far as I can tell, the valid cases are a superset of the supported cases and only weird edge cases (like abbreviated IPv4 addresses - "127.1" etc.) are not validated.

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
```diff
@@ -91,6 +102,34 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
+
+             Valid should be everything accepted by InetAddress.getByName(...),
+             that is: IPv4 addresses, IPv6 addresses, hostnames and FQDNs.
+
+             Additionally, no network requests should be performed as part of
+             the validation, so simply using InetAddress.getByName(...) with a
+             try/catch is out of the question since it might try to resolve
+             domain names.
+             */
+            InternetDomainName.isValid(value) || HostSpecifier.isValid(value)
```
ialokim, line 127: Couldn't we use something like [`Patterns.DOMAIN_NAME`](https://developer.android.com/reference/android/util/Patterns#DOMAIN_NAME) and `Patterns.IP_ADDRESS`? For some reason the current check accepts "g" as host for me which doesn't make too much sense I'd guess

app/src/main/java/de/grobox/transportr/utils/TransportrUtils.kt
```diff
@@ -114,3 +119,16 @@ object TransportrUtils {
 }
 
 fun Location.hasLocation() = hasCoord() && (latAs1E6 != 0 || lonAs1E6 != 0)
+
+fun updateGlobalHttpProxy(newProxy: Proxy, manager: TransportNetworkManager) {
```
ialokim, line 123: Might be better to move this inside the `TransportrUtils` object (static class).

ialokim (changes requested):
Thanks for adding the validation part!

I've just tested your changes inserting an invalid hostname (in this case only the letter "g"). The validation doesn't complain about this (strange enough?), but also no other feedback is given that the proxy connection could not be established. I was able to make network requests as before over the normal internet connection without further notice.

I think the best thing would be to somehow test the proxy settings by performing a request after leaving the settings screen and report invalid values to the user straight away. At least no further network requests should be sent through the normal connection without informing the user.

T0astBread:
Finally another update from me! Proxy connections are tested after the user changes a proxy preference now. If the test fails, an AlertDialog with the error message is shown. Preferences are still accepted on a failed test since it would otherwise be impossible to change e.g. the host and port at the same time.

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
T0astBread: Thanks, I didn't know about those. Although "g" should theoretically be okay since it's a valid hostname and it could theoretically also resolve. This is also how localhost works. (And "localhost" is also not accepted with Patterns.)

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
```diff
@@ -91,6 +103,37 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
+                pref.setOnPreferenceChangeListener { _, newValue ->
+                    try {
+                        val newProxy = settingsManager.getProxy(mapOf(
+                            Pair(prefKey, newValue)
+                        ))
+                        TransportrUtils.checkInternetConnectionViaProxy(newProxy)
+                        TransportrUtils.updateGlobalHttpProxy(newProxy, manager)
+                    } catch (e: Exception) {
+                        Log.e(TAG, "Invalid proxy settings: " + e.message)
+                        AlertDialog.Builder(context!!)
```
ialokim, line 118: I would prefer a less disturbing Toast message at this point.

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
```diff
@@ -91,6 +103,37 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
+                            .create()
+                            .show()
+                    }
+                    true
+                }
+            }
+        }
+
+        findPreference<ValidatedEditTextPreference>(PROXY_HOST)!!.validate = { value ->
+            Patterns.DOMAIN_NAME.matcher(value).matches() || Patterns.IP_ADDRESS.matcher(value).matches()
```
ialokim, line 131: Perhaps we should add "localhost" then manually as I would see it as a common setting?

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
```diff
@@ -92,6 +100,23 @@ class SettingsManager @Inject constructor(private val context: Context) {
             }
         }
 
+    @Throws(UnknownHostException::class, IllegalStateException::class)
+    fun getProxy(proxyPrefOverrides: Map<String, Any>): Proxy {
```
ialokim, line 104: All uses of this function seem to set at most one new setting. Why not simply using a `Pair<String, Any>` here?

ialokim (changes requested):
Sorry for the delay. I've finally had another look at your changes. Please see my comments in-code.

I still think it would be nice to finally test the proxy settings when leaving the settings screen (`onBackButtonPressed` or something similar) and in case the connection could not be established show another Toast message and deactivate the proxy again.

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
T0astBread: You're right but I think the code is actually more readable with the map compared to a single `Pair<String, Any>?`, if we want to handle mismatched types the same way.

Compare
```kotlin
val isEnabled = proxyPrefOverrides[PROXY_ENABLE] as Boolean? ?: settings.getBoolean(PROXY_ENABLE, false)
```
to
```kotlin
val isEnabled =
    if (proxyPrefOverride?.first == PROXY_ENABLE && proxyPrefOverride.second is Boolean)
        proxyPrefOverride.second as Boolean
    else
        settings.getBoolean(PROXY_ENABLE, false)
```
or (avoiding the explicit cast)
```kotlin
val overrideKey = proxyPrefOverride?.first
val overrideVal = proxyPrefOverride?.second
val isEnabled = if (overrideKey == PROXY_ENABLE && overrideVal is Boolean)
    overrideVal else settings.getBoolean(PROXY_ENABLE, false)
```

T0astBread:
I've avoided automatically deactivating the proxy since I figured someone might not notice the Toast and accidentally use the app without a proxy. If the app just continues using invalid proxy settings (which blocks all connections), there's no way a user won't notice something's wrong.

grote:
I don't have time at the moment to do a review, so just a comment: When setting a proxy, I'd show a progress bar until we know the proxy works and then only set the setting for real when it does. Otherwise don't set the settings and show an error dialog.

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
T0astBread: Do you think it would be a good idea to handle some common errors using a Toast with a simplified message and use an AlertDialog with the Exception's message for anything unexpected?

T0astBread:
Where should the progress bar be placed? At the top or bottom of the settings activity, perhaps together with a Snackbar displaying something like "Checking proxy settings..."? Or a progress dialog (those are deprecated)? And what should happen when the user exits the settings activity while the proxy connection is being checked?

I could add another set of proxy preferences that act as "untested" proxy preferences, which are what the user actually sees in the UI. Then, when the user exits the settings activity and the untested settings differ from the currently used ones, the untested settings are checked with a non-blocking progress bar. If the user presses back to exit the activity a Toast shows up saying something like "Press back again to cancel the proxy check". Something similar is probably also possible for the back arrow in the app bar.

But is that really necessary? I know, blocking the UI for a network request is really bad UX but who, other than power users, would mess with proxy settings anyways?

grote:
The network request would have to be done off the UI thread of course, but the UI could still artificially be blocked. Just show a Dialog fragment with a progress bar and a text saying "testing proxy settings". the same fragment could even show the error message and attach to the same ViewModel as the settings.

ialokim:
I've just have done some comparison on how other apps handle proxy settings:
- F-Droid and Aurora Store simply crash after changing to invalid proxy settings, I had to clear the application data to get them to work again - definitely not the way to go.
- DAVx5 at least doesn't crash, but simply reports a general network error as it would do whenever the Cal/CardDav server would not be reachable.
- Telegram FOSS is the only app I've tested that actually gives the user some feedback directly to the proxy settings. See the following screen record:

https://user-images.githubusercontent.com/18088991/108270436-bceeb080-716f-11eb-9632-d99e4e5e0bda.mp4

I think it would make sense to move the proxy settings to a new "sub-preference" screen similar to how Telegram does it. Of course I would argue one proxy setting at a time should be enough. But then we could really check the proxy settings before returning to the main preference screen and prevent "false" or non-working settings. Also, in the main preference screen, there could be a description field stating whether the proxy settings are working and enabled or not.

T0astBread:
So, something like this?

Diagram describing user interface states: A settings screen with a link to a detailed "Proxy Settings" screen. The proxy settings screen checks, upon clicking a "save" check mark, the entered proxy settings and either returns the user to the previous screen or displays an error. The settings screen also indicates if the proxy settings work with a non-blocking connection status label.

ialokim:
Yes, that looks really good to me! Just two minor things:
- Perhaps we could combine the switch and the area to go to the proxy settings into one single row, as e.g. in Android system's settings for Wifi. See [this](https://stackoverflow.com/questions/50433203/how-to-properly-customize-switchpreferencecompat-so-that-it-behaves-like-wi-fi-s) for an image, but not sure if the answer helps much. I've also found [this](https://gist.github.com/marchold/45e22839eb94aa14dfb5) but I could imagine that something similar might exist natively by now.
- The check should also be triggered when going back from the proxy settings to the main settings using the back arrow. Not sure if a dedicated check mark in the top right would be necessary then.

grote:
@T0astBread thanks for the mockup, this looks great!

I personally don't think that we need these colored status information on the main settings screen for proxy settings. Just say if a proxy is set or if it isn't.

If we allow the user to leave the proxy screen while still checking the settings, they either won't see the result (success or fail) or we don't set the  proxy settings at all. The first option is probably better, but then it might be overkill to add checking status info to the main screen. Just update the proxy livedata when we have the result.

Issue #728, opened 2020-11-21
Add support for address formats in different locales (ie North American where the number of a building precedes the street name)


AcharyaS97:
**Acknowledgments**
Please check the following boxes with an `x` if they apply:
* [x] The feature I want to propose would be useful for the majority of users, not only for me personally.
* [ ] I am aware that Transportr is mostly developed by one person in their unpaid spare time.
* [x] I can help myself to get this feature implemented or know someone who wants to do it.
* [ ] If I want to add support for a new region or country, I checked that this is already available in [public-transport-enabler](https://github.com/schildbach/public-transport-enabler/) and know the process [described on the Transportr homepage](https://transportr.app/contribute/#translating-into-your-language).

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. E.g. I'm always frustrated when [...]

Something I've noticed is that the address formats when searching for locations, or saved (ie Home/Work) always seemed a bit off. The street name always came before the number (ie Pennsylvania Avenue, 1600, Washington DC) . Being an ignorant Canadian, I've only ever seen addresses written the other way around(1600 Pennsylvania Avenue, Washington, DC), I thought it looked off. I figured it was a European thing but even between countries it seems like the standard differs:

https://www.parcelforce.com/help-and-advice/sending/address-formats-europe

**Describe the solution you'd like**

An option in app settings where you can pick which address format you want? Different locales might have different conventions which can be customized towards later but as a start just a simple radio button selection in settings where you can pick "Address Format 1" vs "Address Format 2" and have the app display all address locations with either street name first or building number first?

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions you've considered. But please limit this ticket to one feature or use-case.

The only other option I do have is to deal with it and get used to the address format as it is now ;)

**Additional context**

I'm interested in contributing to this project and would love to tackle this addition on my own, if it is deemed worthy of potentially adding.

ialokim:
#559 might be of interest to you. You would have to investigate where the shown order of the address is composed in Transportr. In the Navitia data used as a data source, the `label` field already adheres to the right order (but not so for e.g. Vienna, Austria): http://canaltp.github.io/navitia-playground/play.html?request=https%3A%2F%2Fapi.navitia.io%2Fv1%2Fcoverage%2Fus-wa%2Fplaces%3Fq%3D1600%2520Pennsylvanian%2520Avenue%26type%255B%255D%3Daddress%26

[...]

PR #731, opened 2020-12-15
Issue #728 - Add support for address formats in different locales

AcharyaS97:
Minimal, initial fix for issue #728. We add an enum to the TransportNetwork to specify the output format (Street Name first vs Street Number First) and use that to set the format in the LocationAdapter, through a LocationFormatChanger object.

Idea was to intercept search results during the filter phase and use the LocationFormatChanger object to modify the format of the search result string.


AcharyaS97:
I accept the CLA

AcharyaS97:
I accept the CLA ✔️

grote:
Thanks for the PR! The clabot will only shutup once the commit with the CLA acceptance has been merged to master unfortunately.

AcharyaS97:
No worries. So what do you think about my set of changes?

app/src/main/java/de/grobox/transportr/locations/LocationAdapter.java
```diff
@@ -218,4 +221,9 @@ void setSort(FavLocationType favLocationType) {
 		resetDropDownLocations();
 	}
 
+	void setLocationFormatChanger(TransportNetwork.LocationFormat format){
```
ialokim, line 224: Naming: You're not setting the changer, but the locationFormat here.

app/src/main/java/de/grobox/transportr/locations/LocationFormatChanger.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        //Keep appending tokens until you get to the very last token that you already inserted
+        for (i in 0 until tokens.size - 1) {
+            str.append(tokens[i])
+            str.append(" ")
+        }
+
+        return str.toString().trim { it <= ' ' }
+    }
+    private fun isInteger(s: String): Boolean {
```
ialokim, line 78: Might be checked in a more idiomatic way using [`toIntOrNull`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/to-int-or-null.html).

ialokim (changes requested):
Thanks for your PR!

This only influences the suggestions as displayed in the search right? Would it make sense to change formatting also at other places, e.g. in the trips view / work/home locations? (Or are addresses already correctly formatted there?) If not, I fear you might get inconsistent names throughout the app.

app/src/main/java/de/grobox/transportr/locations/SuggestLocationsFilter.java
```diff
@@ -32,16 +33,16 @@
 	/**
 	 * Starts an asynchronous filtering operation on a worker thread.
 	 */
-	FilterResults performFiltering(CharSequence constraint, List<WrapLocation> favoriteLocations, List<SuggestedLocation> suggestedLocations) {
+	FilterResults performFiltering(CharSequence constraint, List<WrapLocation> favoriteLocations, List<SuggestedLocation> suggestedLocations, LocationFormatChanger locationFormatChanger) {
 		FilterResults filterResults = new FilterResults();
 		if (constraint == null) return filterResults;
-		List<WrapLocation> result = new ArrayList<>();
+		HashSet<WrapLocation> result = new HashSet<>();
```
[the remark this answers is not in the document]
AcharyaS97: Oh that's true. I could use a LinkedHashSet to maintain both order and uniqueness?

app/src/main/java/de/grobox/transportr/locations/LocationFormatChanger.kt
AcharyaS97: Sure!

app/src/main/java/de/grobox/transportr/locations/LocationAdapter.java
AcharyaS97: Right. I will change the name

app/src/main/java/de/grobox/transportr/locations/SuggestLocationsFilter.java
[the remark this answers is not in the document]
ialokim: Sounds good to me.

grote:
AFAIK this is a problem of Navitia and should be addressed there. When I met them some years back in Paris, they already said they will be launching a new search engine that handles different address format better. I wonder what become of this.

AcharyaS97:
The reason I figured it would be alright leaving the updates out in other places was from the perspective that people wouldn't necessarily change TransportNetworks all too often and if they did, they wouldn't care about the format olocations they saved in other those places (ie work or home or favourite trips) . If you're based in an area where the address is in one style and your preferences are in that style then you go travelling and you switch transport networks, you wouldn't really care if it changed the locations you saved in other networks. I guess that leads into being able to set an address format as a setting though.

Also to the other point, I agree its an upstream issue but lieu of a solution there I figured it might improve usability of the app immediately for people in the jurisdictions of transport networks like mine where the upstream data format isn't the same as what the "accepted" format is.

ialokim:
> The reason I figured it would be alright leaving the updates out in other places was from the perspective that people wouldn't necessarily change TransportNetworks all too often and if they did, they wouldn't care about the format olocations they saved in other those places (ie work or home or favourite trips) . If you're based in an area where the address is in one style and your preferences are in that style then you go travelling and you switch transport networks, you wouldn't really care if it changed the locations you saved in other networks. I guess that leads into being able to set an address format as a setting though.

You might have misunderstood what I meant: Even while using the same TransportNetwork, you have several places in the app where addresses might be shown - the search suggestion pop-up is only one of them. Others include addresses in routing results or the work/home location set for this TransportNetwork. I guess it would also be necessary to post-process the addresses there?

> Also to the other point, I agree its an upstream issue but lieu of a solution there I figured it might improve usability of the app immediately for people in the jurisdictions of transport networks like mine where the upstream data format isn't the same as what the "accepted" format is.

I agree that it makes sense to improve usability for users if the upstream is not going to fix it soon. Perhaps it might be good to ask them before, how their plans are concerning the new search engine. At last, this "hack" should be clearly marked as such and the respective functions should only be called for the specific TransportNetworks where necessary.

Issue #732, opened 2020-12-16
Swapping From and To fields sometimes results in an animation glitch


AcharyaS97:
**Describe the bug**

It looks like there's a bug in the DirectionsFragment where if the From and To fields are both empty and you try swapping them,  there's an animation glitch where it'll do the swap but then the hint text will reset. I saw that there have been a few issues relating to the DirectionsFragment with the 2 LocationViews for From and To, and there was a Pull Request  #722 that addressed some problems. If that PR or some other PR addresses the problem I mention here then do let me know. 

**To Reproduce**

(2a and 2b are 2 separate usecases where I saw the defect occur.)

1. Open the directions view (click the left turn arrow in the map view)

2. 
a) Make sure that both the From and To fields are empty and only the hint text is showing
b) Type something in either From or To but don't select an option from the search results

3. Click the swap button on the UI.

Notice that if the fields are empty, the swap will happen but the hint text on the From and To fields will reset to their original positions after the animation ends. If either field is populated but it's not a result from Search task, that field will be erased.

**Expected behavior**

For the 2a case (with both fields empty), the simplest solution might be just to not do anything if the swap is clicked and both From and To fields don't have any text in them. 

For the 2b case (where the fields are populated with text but don't have a Location selected from the search results) I think it might be a little more complicated.

From looking at how the swap works, I understand that it doesn't actually change the positions of the two LocationViews, just swaps the underlying saved locations in those views. We could update this to actually swap the positions of those views? Alternatively we could change to swap the text of the LocationViews by allowing the option to set the Text of the LocationView independent of the actual Location object that it has?

ialokim:
Thanks for your issue report! These are indeed independent of #722.

I would propose to just disable/hide the button in both cases as an easy bugfix.

AcharyaS97:
Alright, I'll update it!

AcharyaS97:
Also saw a similar issue with the "Favourite" icon where if you had valid From and To locations, but then you used the clear button on either one, the "Favourite" icon would not go away, which doesn't make sense. Have pushed fix for both issues in associated PR.

PR #733, opened 2020-12-17
Fix issues with icons in DirectionsFragment 

AcharyaS97:
Simple fixes for issues seen in DirectionsFragment having to do with the Swap and Favorite Trip Icons

Only show the "Favourite Trip" and "Swap" icons when both LocationViews are populated.

Refer to Issue #732


ialokim:
Thanks for your PR! I'd say the conceptually cleaner way would be to not rely on the LocationView's `getLocation()`, but add a [LiveData Transformation](https://developer.android.com/reference/androidx/lifecycle/Transformations) inside the ViewModel and only observe and react to the changes inside the Fragment.

grote:
The Favourite Trip Icon does not relate to the locations, but to the displayed trips and should only be displayed when trips are displayed.

grote:
I vaguely remember some design guidelines that say you should disable icons rather then hide them when they are not available.

ialokim:
> The Favourite Trip Icon does not relate to the locations, but to the displayed trips and should only be displayed when trips are displayed.

Then it might make sense to hide the trip results when the from or to field is cleared?

grote:
Maybe yeah. I think I might have left it until the next search in case the user still needs it.

ialokim:
> I vaguely remember some design guidelines that say you should disable icons rather then hide them when they are not available.

Is there a nice-looking and understandable way of disabling borderless icon buttons in the app top bar?

AcharyaS97:
Thank you for all the feedback! Addressing specific points,

> Thanks for your PR! I'd say the conceptually cleaner way would be to not rely on the LocationView's getLocation(), but add a LiveData Transformation inside the ViewModel and only observe and react to the changes inside the Fragment.

Sure, so instead of putting the ui state change code into the from/to location observables, create a new observable for the state of the swap icon that derives its value from the state of the from/to observables in the viewmodel through a Transformation?

> The Favourite Trip Icon does not relate to the locations, but to the displayed trips and should only be displayed when trips are displayed.

Right ok. So I should tie the state of that Favourite Icon to when the trips member in the ViewModel is updated?

> Then it might make sense to hide the trip results when the from or to field is cleared?

In my opinion as a user, this would be fine to leave as is in case it was cleared by accident or if the user still needed the result of that last trip query to decide on what their next query would be?

ialokim:
> Sure, so instead of putting the ui state change code into the from/to location observables, create a new observable for the state of the swap icon that derives its value from the state of the from/to observables in the viewmodel through a Transformation? What's the Transformation of types that would be occurring?

I'd say the location types (from/to) -> boolean (show or hide swap icons).

> So I should tie the state of that Favourite Icon to when the trips member in the ViewModel is updated?

I think this is already how it is done currently.

AcharyaS97:
https://user-images.githubusercontent.com/10468158/103258651-fd962e80-4963-11eb-991e-4202c7aa022a.mp4

I've updated it to be more idiomatic as requested. Also about disabling the icon instead of hiding it, refer to the video clip. As per material design guidelines, disabled icons should have an opacity of 38% of the regular icon if it is black. Because the starting icon is white, I went the other way, setting the disabled color to #999999 compared to the #FFFFFF of the enabled variant. If you like I can apply the same styling to the other elements as well?

ialokim:
Thanks for your continuous work! To be honest, I'm still not really convinced by the disabled icon look. This grayish color looks a bit odd and unmotivated to me. Not sure whether @grote has another opinion or an idea on how to make this more understandable and perhaps nicer looking?

AcharyaS97:
I've been playing around with it. What if the icon itself was changed to something like swap_vertical_circle, since it has more surface area, and the opacity was reduced on disable.

Enabled

Disabled

grote:
I agree that the first version doesn't look nice. The second is already better, but ideally we wouldn't need to change the icon.

Also, did someone check how normal action bar icons look when they are disabled? Maybe we can re-use that style somehow?

AcharyaS97:
I was thinking that the new one might be easier to see when disabled compared to the current icon, but I'll change that back if you like. Also, did you mean a style that's used in the app elsewhere, or the general design guideline for styling of a disabled icon in an action bar? 

Also, final thing but is the behaviour in the clip working as expected? If both from/to are populated and the back button is pressed, should they be cleared as well?

https://user-images.githubusercontent.com/10468158/103427286-f8b4c300-4b8d-11eb-8d31-27b47614bf1c.mp4

PR #735, opened 2020-12-23
Hide via button if current network provider does not support it

### LINKED ISSUE #711: Hide Via toggle in directions form if not supported
After the first part of https://github.com/schildbach/public-transport-enabler/pull/333 has been merged in, we should use this information for hiding the "Via" toggle for providers that don't support this (mainly Navitia-based ones).

ialokim:
Fixes #711.

### DIFF
--- a/app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
+++ b/app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
@@ -39,6 +39,7 @@ import de.grobox.transportr.utils.DateUtils
 import de.grobox.transportr.utils.LiveTrigger
 import de.grobox.transportr.utils.SingleLiveEvent
 import de.schildbach.pte.NetworkId
+import de.schildbach.pte.NetworkProvider
 import de.schildbach.pte.dto.Product
 import de.schildbach.pte.dto.Trip
 import java.util.*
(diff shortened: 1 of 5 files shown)


app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
```diff
@@ -207,5 +209,6 @@ class DirectionsViewModel @Inject internal constructor(
             application.applicationContext, network.networkProvider,
             settingsManager, locationRepository, searchesRepository
         )
+        viaSupported = MutableLiveData(network.networkProvider.hasCapabilities(NetworkProvider.Capability.TRIPS_VIA))
```
grote, line 212: If this is static information known at the time of `init` and never changing, it wouldn't need to be a LiveData I guess, but its fine.

grote (approved):
Great work. I wanted to have this for a long time already.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
ialokim: You are right, it would not be necessary. But I guess it won't harm either and the advantage is to have only LiveData in the interface to the ViewModel.

grote: I don't see this as a big advantage, but sure it doesn't hurt.

PR #744, opened 2021-02-20
Fix German translation

pjaap:
Found a translation mistake.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -16,7 +16,8 @@
     "hoejmann",
     "brandsimon",
     "sideeffffect",
-    "T0astBread"
+    "T0astBread",
+    "pjaap"
   ],
   "label": "cla-signed ✔️",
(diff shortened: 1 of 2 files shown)


ialokim:
Thanks, good spot! We are managing the translations for Transportr as a Transifex project. If you would be interested in joining the German translation team, feel free to sign up there. More information can be found on [our website](https://transportr.app/contribute/#translating-into-your-language).

If, however, you just want to get this little fix included in Transportr and don't want to get a new account at another service, I could change the translation at Transifex. Just report back what you would prefer.

ialokim:
Btw, I even would prefer "wenigste Umstiege" instead of "geringste".

pjaap:
Hi! Thanks for the feedback. I registered at Transifex (with my github profile), but cannot join the "German" Transportr Team. Do I need a confirmation?

ialokim:
I've just accepted you.

pjaap:
Thanks! I *think* I fixed it now in Transifex :)

PR #750, opened 2021-05-18
Fixed function definition to actually override onSaveInstanceState

fgndev:
I am building with rather recent tools, guess that's why this bug showed up. Fixed it.
https://developer.android.com/reference/kotlin/android/app/Activity#onSaveInstanceState(android.os.Bundle,%20android.os.PersistableBundle)


fgndev:
I accept the CLA

ialokim (approved):
It doesn't change the behavior, but as the java function is annotated with [`@RecentlyNonNull`](https://android-developers.googleblog.com/2020/03/handling-nullability-in-android-11-and.html) and according to the doc that you've linked, I think we should indeed correct that. Thanks!

Issue #533, opened 2018-10-15
Add option for "maximum walking time"


bisam:
It would be nice to be able to allow more walking time (or less) then the default settings allow.

Other apps have a slider for it.

Also: Thanks for this amazing app! :)

grote:
Hey thanks for the proposal!

Unfortunately, the walking speeds are fixed in the library Transportr uses. So unless someone adds more fine-grained values to the library, Transportr can't do anything here: https://github.com/schildbach/public-transport-enabler/blob/master/enabler/src/de/schildbach/pte/NetworkProvider.java#L60

PR #752, opened 2021-05-22
Hide walking route unless it's the only one

starsep:
# Rationale
I use the app mostly in the city I live in (Warsaw, Poland) - quite natural use case.
I know the city good enough that I don't really look for the public transport route if I would prefer to walk instead.
Therefore walking route is pretty much never a viable alternative - it's distracting and annoying to see it.

One possible solution would be to have "maximum walking time" as in #533 
I decided to implement something simpler which works for me: just hide the walking route if it's not the only one.

# Screenshots
Quite extreme case obviously - route between two subway stations, 11 minutes vs an hour. With the setting turned off and on.
image
image

Settings screen
image

PS I haven't created issue because I don't think it's "major work" but I can create one if needed.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -17,7 +17,8 @@
     "brandsimon",
     "sideeffffect",
     "T0astBread",
-    "fgndev"
+    "fgndev",
+    "starsep"
   ],
   "label": "cla-signed ✔️",
   "message": "Thank you for your pull request and welcome to our community! We require contributors to sign our [Contributor License Agreement](https://github.com/grote/Transportr/blob/master/CLA.md), and we don't seem to have the user {{usersWithoutCLA}} on file. In order for your code to get reviewed and merged, please explicitly state that you accept the agreement. Alternatively, you can add a commit that adds yourself to https://github.com/grote/Transportr/blob/master/.clabot"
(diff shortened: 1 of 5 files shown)


starsep:
I accept CLA

ialokim:
Thanks for your contribution!

> I accept CLA

No worries, the bot will be happy once `.clabot` is on master.

@grote, what do you think of this additional option in the settings screen? I see their point.

grote:
I agree with the rationale of this PR, but I'd wished we'd talked about this before putting in the work, because I don't think it is Transportr's job to manually change provider results like this. Instead, this should be done in PTE, or even better in Navitia which is providing these strange results. Maybe they even already have query parameters that could be added to PTE to avoid these routes there.

starsep:
> I agree with the rationale of this PR, but I'd wished we'd talked about this before putting in the work, because I don't think it is Transportr's job to manually change provider results like this. Instead, this should be done in PTE, or even better in Navitia which is providing these strange results. Maybe they even already have query parameters that could be added to PTE to avoid these routes there.

Frankly it was easier to me to just implement this feature than wait for discussion, especially as I didn't want this annoyance to stop me from using Transportr.
I agree it should be handled by data provider if possible. I will check PTE/Navitia and see what can be done there, thanks for the great app!

starsep:
In the end I stopped using Transportr and didn't follow up in PTE. I think it's useful feature

Altonss:
> In the end I stopped using Transportr and didn't follow up in PTE. I think it's useful feature

No worries :) I haven't used Transportr too much lately too, but hope to be able again if development picks up speed again! We might discuss this feature in the future again (maybe also opening an issue in PTE), after more urgent issues are taken care of.

Issue #693, opened 2020-08-19
Switch to standard Gradle dependency verification


ialokim:
As a followup of #692, we might consider switching dependency pinning from https://github.com/akwizgran/gradle-witness to the new standard Gradle one: https://docs.gradle.org/current/userguide/dependency_verification.html.

Refer to https://gitlab.com/fdroid/fdroidclient/-/commit/be5bdf32198f674563b30c84a57fb947bdaf33da and https://gitlab.com/fdroid/fdroidclient/-/commit/708a6d8dbdb55a55c7864981c4e10587bec36c37 as an example.

PR #759, opened 2021-06-01
Built-in gradle dependency verification

ialokim:
As discussed in #693, this PR switches from the gradle witness plugin to the built-in [gradle dependency verification](https://docs.gradle.org/current/userguide/dependency_verification.html). As before with the gradle witness approach, only the checksums of the dependencies are verified, not the GPG signatures. However, the metadata files (`.pom`) are now verified, too, and locally built dependencies are ~(should be)~ already automatically excluded. The only platform-dependent dependency is `aapt2`, I had to add the checksums for the windows and osx version manually. ~Let's see if the verification works properly in the CI~ As the verification worked in the CI, this should be ready to be merged :smile:


update-dependency-pinning.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 
-WITNESS=app/witness.gradle
+# this will just append new checksums and not remove unnecessary ones
 
-echo "" > $WITNESS
-./gradlew -q calculateChecksums | grep -Ev "^(Skipping|Verifying)" | grep -Ev "files-2.1:|caches:transforms-3:|:build-tools:core-lambda-stubs.jar:|:platforms:android.jar:|-linux.jar:" > $WITNESS
+# to clean up the file, remove ./gradle/verification-metadata.xml,
+# run the command below and manually (re-)add checksums for missing operating systems windows, osx or linux for aapt2
+# checksums can be computed after downloading the respective jars of https://maven.google.com/web/index.html?q=aapt2#com.android.tools.build:aapt2
```
grote, line 7: Its a pity we can't do that automatically. Would be nice to have that file reflect all changes and not grow bigger over time.

update-dependency-pinning.sh
```diff
@@ -1,7 +1,9 @@
...
 
+./gradlew --write-verification-metadata sha256 build
```
grote, line 9: Does this overwrite existing entries? What happens if a checksum has changed, e.g. now it gets a different one than was in that file before?

update-dependency-pinning.sh
ialokim: I agree, that would be better, but only adding new entries avoids removing manually added checksums as the ones for windows and osx for `aapt`. I think it will make sense anyway to look at the diff after running this command, so perhaps it is viable to remove old versions manually then. After all, this should be only updated after library upgrades.

update-dependency-pinning.sh
ialokim: I am not fully sure. I suspect it would add the new checksum as a second option. Looking at the diff will again reveal the behavior and one could manually react to it.
Btw, the gradle witness plugin would silently change the checksum and one would only notice it looking at the diff, too.

update-dependency-pinning.sh
grote: >  only adding new entries avoids removing manually added checksums as the ones for windows and osx for aapt.

but it won't help after we upgrade the gradle plugin as this will pull in new dependencies again that we would need to add manually, right?

> perhaps it is viable to remove old versions manually then

maybe easier to re-add windows/mac versions if we even want to support those?

ialokim: > > only adding new entries avoids removing manually added checksums as the ones for windows and osx for aapt.
> 
> but it won't help after we upgrade the gradle plugin as this will pull in new dependencies again that we would need to add manually, right?

That's correct.


> > perhaps it is viable to remove old versions manually then
> 
> maybe easier to re-add windows/mac versions if we even want to support those?

We could certainly change the bash script to always remove the old verification data (as was done with the witness plugin before), but I would have to check if the order of the dependencies is deterministic to ensure sensible diffs.

goapunk: According to https://github.com/gradle/gradle/issues/11664 the entries are sorted, so diffs should be fine

PR #760, opened 2021-06-03
Add Washington DC Provider

liamnorm:
Adds support for Washington, DC using the "us-dc" location from Navitia.


app/witness.gradle
```diff
@@ -84,6 +84,7 @@ dependencyVerification {
         'com.android.tools.analytics-library:shared:27.1.3:shared-27.1.3.jar:10d2a51d8f89ff4ac849888e5a9c60b10e879c30d78545ec1da4d3df7bd56ae4',
         'com.android.tools.analytics-library:tracker:27.1.3:tracker-27.1.3.jar:589b355a2ba796cbc0a2b2295737de6661f078262e5f87cd6f540b8d011e5ebb',
         'com.android.tools.build:aapt2-proto:4.1.0-alpha01-6193524:aapt2-proto-4.1.0-alpha01-6193524.jar:17e75523e1e92dd4f222c7368ee41df9e964a508232f591e265d0c499baf9dca',
+        'com.android.tools.build:aapt2:4.1.3-6503028:aapt2-4.1.3-6503028-osx.jar:6c57e78049706e57ff65ad8c5fef43ed840c086710aeafa701953d6abbc90724',
```
ialokim, line 87: This is problematic as it is platform-dependent. Will have to remove that before merging. If #759 gets merged first, that should be no problem.

app/src/main/res/values/strings.xml
```diff
@@ -422,6 +422,8 @@ and always knows where you are to not miss where to get off the bus.
 	<string name="np_desc_cmta">Austin</string>
 	<string name="np_name_usny">New York and New-Jersey States</string>
 	<string name="np_desc_usny" translatable="false">MTA, CDTA, Port Authority, Staten Island Ferry, NFTA, Rockland County, Rochester-Genesee Regional Transportation Authority</string>
+	<string name="np_name_usdc">Washington, DC</string>
+	<string name="np_desc_usdc" translatable="false">Metro</string>
```
ialokim, line 426: Is it really only metro?

ialokim (changes requested):
Thanks! Looks good overall.

app/src/main/res/values/strings.xml
liamnorm: And Metrobus and Capital Bikeshare (which I just added to the string). I also sent a message to Navitia asking if they could add ART bus data.

Is there a way I could use the Navitia playground to see all the services provided? Right now I'm just sending requests and looking at the results. I can't tell if I have everything.

liamnorm:

app/src/main/res/values/strings.xml
ialokim: Hi @liamnorm, what is your opinion regarding WMATA? Is it worth including in the network description string?

liamnorm: Well, Metro is the same thing is WMATA (the 'M' is for Metropolitan). Everyone I know calls it Metro. I don't think adding WMATA is necessary.

app/witness.gradle
Altonss: Is this dependency needed?

ialokim:
See #801 . Navitia unfortunately stopped serving the data for that region.

PR #801, opened 2022-01-20
add WashingtonDC for UI

1305533243:
added support for the WashingtonDC area

### DIFF
--- a/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
+++ b/app/src/main/java/de/grobox/transportr/networks/TransportNetworks.kt
@@ -654,6 +654,13 @@ private val networks = arrayOf(
                         description = R.string.np_desc_usny,
                         status = ALPHA,
                         factory = { NewyorkProvider(NAVITIA) }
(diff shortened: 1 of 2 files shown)


Altonss:
I would need to test this PR, but if it works it could be integrated fairly easily :)
ping @grote

pt2121:
Thanks for the contribution. Looks like this doesn't work anymore - I ran into this when we call Navitia API for US DC. `https://api.navitia.io/v1/coverage/us-dc/places?q=mal&type%5B%5D=stop_area&type%5B%5D=address&type%5B%5D=poi&type%5B%5D=administrative_region&depth=1`
```
{"error":{"id":"unknown_object","message":"The region us-dc doesn't exists"},"message":"The region us-dc doesn't exists"}
```

I don't think there is a way around this since we rely on the API. I'll close this unless we have a workaround.

ialokim:
Yeah, unfortunately Navitia has removed a lot of regions lately. Let's recompile a list here #877. Closing this since it doesn't work any longer.

PR #814, opened 2022-04-26
Fix not finding installed calendar apps

### LINKED ISSUE #795: Calendar application cannot be found on 2.1.4
**Describe the bug**
Transportr is unable to find my installed calendar app (Etar), which makes adding connections to your calendar impossible.
This only applies to 2.1.4; on 2.1.3 and previous versions it works fine.

**To Reproduce**
1. Select any connection in the app
2. Press the calendar icon in the top right corner

**Expected behavior**
Transportr would open my calendar app to create a new event with all the details of the selected connection

 - **Transportr Version:** 2.1.4
 - **Device:** Pocophone F1
 - **Android Version:** 11 (LineageOS for MicroG 18.1)

### LINKED ISSUE #818: Share button not working
**Describe the bug**
the 'Share' button results in a toast saying 'No external map installed :('

**To Reproduce**
0. install a map app compatible with geo intent (eg. organicmaps, osmand)
1. make a point on the map
2. Click on 'share'
3. See error

**Expected behavior**
it should trigger a geo intent 

**Screenshots**
Screenshot_20220615-192752_Transportr

 - **Transportr Version:** 2.1.4
 - **Device:** Oneplus 8 Pro
 - **Android Version:** 12

ialokim:
handle ActivityNotFoundException instead of manually resolving calendar intent (by default not allowed on API level >= 30)

bug reproduced on Android 12 emulator, fix tested on both API 31 and 22 (Android 5.1) with and without calendar applications installed

fixes #795, fixes #818

### DIFF
--- a/app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
+++ b/app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
@@ -20,10 +20,13 @@
 package de.grobox.transportr.utils
 
 
+import android.content.ActivityNotFoundException
+import android.content.ComponentName
 import android.content.Context
 import android.content.Intent
 import android.content.Intent.FLAG_ACTIVITY_CLEAR_TOP
 import android.net.Uri
+import android.os.Build
 import android.util.Log
 import android.widget.Toast
 import com.mapbox.mapboxsdk.geometry.LatLng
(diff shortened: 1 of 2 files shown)


ialokim:
We actually face the same issue with external maps applications that are not found anymore. Perhaps let's hold back this PR until this is also fixed. see #713 as a reference.

app/src/main/java/de/grobox/transportr/utils/IntentUtils.kt
```diff
@@ -100,20 +103,16 @@ object IntentUtils {
...
-            val packageName = it.activityInfo.packageName
-            if (context.packageName != packageName)
-                Intent(intent).apply { `package` = packageName }
-            else null
-        }
-        if (intents.isEmpty()) {
+        val intentChooser = Intent.createChooser(intent,  context.getString(R.string.show_location_in))
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N)
+            // exclude Transportr from list on Android >= 7
+            intentChooser.putExtra(Intent.EXTRA_EXCLUDE_COMPONENTS, arrayOf(ComponentName(context, MapActivity::class.java)))
```
ialokim, line 109: filtering out Transportr from the intent chooser now only works on Android 7+. On older versions, we have the small regression that Transportr is shown again as an option in the chooser (same as before #713), but it simplifies the code a lot.

ialokim (commented):
Just added a second commit fixing the same problem for sharing a location. Successfully tested on API 22/Android 5.1 (still works but shows Transportr again as sharing option) and API 31/Android 12 (finds other installed map applications again).

This would be ready for merging.

grote:
Thanks for picking this up! Want to do a PTE update and release commit as well, so we can get this out?

Issue #772, opened 2021-08-17
remember preferred transportation mode ("Verkehrsmittelwahl")


mark-franz:
**Is your feature request related to a problem? Please describe.**
In 90% of all connections I prefer local transport ("Nahverkehr"). I want to exclude high speed trains or other means of long-distance traffic ("Fernverkehr"), because local transport tickets won't apply. 
I want this preference be stored permanently. 
However, currently I always need to un-select high-speed trains ("Hochgeschwindigkeitszug") for any time I closed and re-opened the app, which often happens to me. 

**Describe the solution you'd like**
Please store the preferred transportation mode (choice of transportation type or service) permanentely. 
Especially for the Deutsche Bahn schedules.

ialokim:
This is a slightly modified version of #484 (where storing the means of transport per network provider is requested) and also related to #635. Especially https://github.com/grote/Transportr/issues/484#issuecomment-408094441 has to be taken into consideration.

PR #833, opened 2022-10-10
Remember preferred means of transport

pt2121:
addresses #772

### DIFF
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
@@ -139,13 +139,13 @@ class SettingsManager @Inject constructor(private val context: Context) {
     fun setPreferredProducts(selected: Set<Product>) {
         val editor = settings.edit()
         Product.ALL.toSet().forEach { product ->
-            editor.putBoolean(product.name, product in selected)
+            editor.putBoolean(LAST_PRODUCT_PREFIX + product.name, product in selected)
         }
         editor.apply()
     }
 
     fun getPreferredProducts(): Set<Product> {
-        val firstTime = Product.ALL.none { settings.contains(it.name) }
+        val firstTime = Product.ALL.none { settings.contains(LAST_PRODUCT_PREFIX + it.name) }
         if (firstTime) {
             setPreferredProducts(Product.ALL)
             return Product.ALL
@@ -153,7 +153,7 @@ class SettingsManager @Inject constructor(private val context: Context) {
 
         val products = mutableSetOf<Product>()
         Product.ALL.toSet().forEach { product ->
-            if (settings.getBoolean(product.name, false)) {
+            if (settings.getBoolean(LAST_PRODUCT_PREFIX + product.name, false)) {
                 products.add(product)
             }
         }
@@ -172,6 +172,7 @@ class SettingsManager @Inject constructor(private val context: Context) {
         private const val OPTIMIZE = "pref_key_optimize"
         private const val LOCATION_ONBOARDING = "locationOnboarding"
         private const val TRIP_DETAIL_ONBOARDING = "tripDetailOnboarding"
+        private const val LAST_PRODUCT_PREFIX = "pref_key_last_product_prefix_"
     }
 
 }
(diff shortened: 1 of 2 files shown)


pt2121:
I accept the [Contributor License Agreement](https://github.com/grote/Transportr/blob/master/CLA.md).

MugguGubbl:
Please consider merging this PR for next release after maintenance release 2.1.5.

I've been using Transportr for years now and waiting for this feature so long, as I never use high-speed trains. So it's really annoying to change the means of transport every time. I think many more people have the same issue since there is the new "Deutschlandticket".

As I'm not a developer I can't help at the restart for the app, that's currently going on. But I made a donation. Hope this helps. BTW I registered for GitHub just to write this ;)

Altonss:
> Please consider merging this PR for next release after maintenance release 2.1.5.
> 
> I've been using Transportr for years now and waiting for this feature so long, as I never use high-speed trains. So it's really annoying to change the means of transport every time. I think many more people have the same issue since there is the new "Deutschlandticket".
> 
> As I'm not a developer I can't help at the restart for the app, that's currently going on. But I made a donation. Hope this helps. BTW I registered for GitHub just to write this ;)

Thanks for the comment :) I hope we can get this merged too, as probably a lot of people need this :)
But it still needs some testing, and maybe some feedback by @ialokim and @grote :)
And also thanks a lot for the donation, I'm sure its much appreciated :)

ping @pt2121 do you want to resolve conflicts on .clabot yourself or I can do it through the github UI?

pt2121:
I can take a look after the release. It's been a few months since I worked on this. I have to retest it a bit.
Thanks for the donation :moneybag:

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
```diff
@@ -121,6 +121,7 @@ class DirectionsViewModel @Inject internal constructor(
     fun setProducts(newProducts: EnumSet<Product>) {
         _products.value = newProducts
         search()
+        settingsManager.setPreferredProducts(newProducts)
```
ialokim, line 124: I don't think it's a good idea to update the settings from within the UI thread. Maybe double-check how it is done for other settings?

pt2121: Totally agree with you on this. That said, from what I see, all `SettingsManager` methods are currently called on the main thread. I will create an issue to improve this.

pt2121: Can we leave this as-is?
This shouldn't block the UI thread and, from what I see, other methods in `SettingsManager` are called from the main thread too.

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
```diff
@@ -135,6 +136,30 @@ class SettingsManager @Inject constructor(private val context: Context) {
         return settings.getBoolean(SHOW_WHEN_LOCKED, true)
     }
 
+    fun setPreferredProducts(selected: Set<Product>) {
+        val editor = settings.edit()
+        Product.ALL.toSet().forEach { product ->
+            editor.putBoolean(product.name, product in selected)
```
ialokim, line 142: Maybe it would be actually better to use some common prefix for those setting entries, just to not pollute the setting file too much. Something like `lastProducts-*` maybe? Obviously this would need to be updated below in the other functions as well.

app/src/main/java/de/grobox/transportr/trips/search/DirectionsViewModel.kt
ialokim: Alright, that's good to know! We can leave it as-is in that case I'd say!

ialokim (changes requested):

app/src/main/java/de/grobox/transportr/settings/SettingsManager.kt
pt2121: Sounds good. I will update :+1:

pt2121:
I will merge this in a week unless someone wants to give more feedback.

ialokim (approved):
LGTM now. Haven't tried it in the app myself though.

PR #835, opened 2022-10-16
Remove Mapbox logo in trip map

Altonss:
I didn't found the mapbox logo in drawables, but maybe it would need to be removed to clean up the project :)
ping @ialokim


Altonss:
ping @ialokim :)

grote:
We had no logo before but got threatened by Mapbox about it. It can go away when we use maplibre maps.

ialokim:
Also whenever we remove the Mapbox logo after changing to MapLibre, we should clean up the corresponding resource and artwork, I guess.

Since this is not applicable until #668 is done and the change is minor anyhow, I will close this PR for now. Feel free to reopen as soon as it applies.

PR #850, opened 2023-04-11
Add Themed Icons

### LINKED ISSUE #837: Add material you icon support

thgoebel:
This PR adds support for Themed Icons (aka monochrome, aka Material You) to the app icon on the launcher.

See also: https://developer.android.com/develop/ui/views/launch/icon_design_adaptive

Tested only on vanilla Android 13 (LineageOS 20), but should (tm) introduce no regression on older Android versions.
Notably, it works on Android 13 (SDK 33) even while still targeting SDK 30.

Fixes #837.

### DIFF
--- a/.clabot
+++ b/.clabot
@@ -18,7 +18,8 @@
     "sideeffffect",
     "T0astBread",
     "fgndev",
-    "Altonss"
+    "Altonss",
+    "thgoebel"
   ],
   "label": "cla-signed ✔️",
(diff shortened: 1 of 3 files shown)


Altonss:
@grote or @ialokim, this seems to be an easy fix/improvement. Would you mind if we merge this for 2.1.5?

ialokim:
> this seems to be an easy fix/improvement. Would you mind if we merge this for 2.1.5?

Looks good for me, but I would like to have someone test this on older Android versions (preferably including the minimum we support) to make sure we don't introduce regressions. Installing once on an emulator and checking the app icon should be fine.

mueller-ma:
I can confirm that this doesn't introduce a regression on an Android 9 emulator.

Issue #852, opened 2023-05-13
The Future of the Transportr App (?)


Echolon:
Dear Transportr Developers & Maintainers,
Dear Transportr Community,
Dear everyone interested,

I would like to raise a general topic concerning the future of this actual beautiful and very useful app that used to get me home and to my points of interest beyond my home-zone and in many countries for years.

To my understanding the development has very much declined if not stopped about [two years ago](https://github.com/grote/Transportr/commits/master) :-(
That's sad, but I also understand that there are good reasons behind it. So far I understood there was
- lack of interest from user side
- lack of support from community
- which results in a lack of motivation
- other projects I assume

Please correct me if I am wrong in any way. Of course, no blame or offense!

I would love to support changing the situation. I am not a developer per se, but I am certainly good at setting and supporting a new framework and conditions how we as a community could set a new direction of Transportr. Moreover, I see that there are also certainly [people try to up-streaming pull requests](https://github.com/grote/Transportr/pulls).

**Would you be interested? Is that something feasible? What could we do?**
I think the first steps could be to summarize the current shortcomings and get a clear picture of hurdles the development experience.

To my superficial understand I see the following points to tackle:

- Discuss if there is interest on a way forward
- If so, spread the word that we are try to start over again
- Review communication and promotion of the app
  - I believe the outstanding good idea of this software has received way to few attention in general. Communication through the Transportr webpage, [Twitter](https://twitter.com/TransportrApp) and maybe even Mastodon chould be reconsidered. The community could certainly help to upstream input here.
- Discuss and promote donations
- Consider funds alike the [NLnet](https://nlnet.nl/) or [OpenTech](https://www.opentech.fund/) fund
- Call previous contributors
- Discuss maintenance
- Discuss further development (but this is certainly not a first thing to discuss)

@grote and @ialokim many thanks for your efforts and devlepment of this app! Would you be interested in anyway to support such a discussion and reconsideration of how this could get a new chance? Do you even share my thoughts?

It would be a pleasure for me to serve and organize around this. Projects like the [StreetComplete](https://github.com/streetcomplete/streetcomplete) app managed to build a great serving community around that app.

Looking forward to revive this idea of modern public transport around the world.

Best regards,
echo

________

I know its not usually the best thing to do, but I hope it is okay that I mention previous contributors since 2022. Otherwise, I apologize.

@htgoebel 
@ashtanko
@Altonss 
@pt2121
@nomq1
@mimi89999
@1305533243
@aha999

[...]

PR #853, opened 2023-05-18
update target sdk version to 31

pt2121:
As mentioned in https://github.com/grote/Transportr/issues/852#issuecomment-1551918703, Google Play new requires the target SDK to be at least 31. Also see https://support.google.com/googleplay/android-developer/answer/11926878.

I went through [the doc here](https://developer.android.com/google/play/requirements/target-sdk#pre12) & [here](https://developer.android.com/about/versions/12/behavior-changes-12), and made sure that we updated everything that we should - fortunately it's not that much. From what I see, we don't use any foreground service, notif, BroadcastReceiver so it seems like a smooth update :crossed_fingers:


pt2121:
@cla-bot I accept the CLA.

Altonss:
@cla-bot check

Altonss:
@cla-bot[bot] check

app/src/main/AndroidManifest.xml
```diff
@@ -96,6 +97,10 @@
 			android:name=".about.ContributorsActivity"
 			android:label="@string/drawer_contributors"/>
 
+		<service
+			android:name="com.mapzen.android.lost.internal.GeofencingIntentService"
+			android:exported="true" />
```
ialokim, line 102: Could you explain why this is needed now?

pt2121: Good question. Basically, for Android 12+, any Android components with intent filters require to explicitly declare the `exported`. More info is [here](https://developer.android.com/about/versions/12/behavior-changes-12#exported).
My understanding is that we use this service internally and I found a solution [here](https://github.com/lostzen/lost/issues/272)

Altonss:
Why not target level 33 directly? Is this significantly more work? I looked here https://developer.android.com/google/play/requirements/target-sdk and found that starting of August, 33 will be required by PlayStore

app/src/main/AndroidManifest.xml
grote: Does it work without exporting the service? i.e. setting exported to false

grote (approved):

Issue #738, opened 2020-12-27
UI bug on API level 16


ialokim:
**Screenshot**
![Screenshot_1609011631](https://user-images.githubusercontent.com/18088991/103171915-0c8bbc80-4850-11eb-9e2a-6e4c5619079b.png)

**Versions (please complete the following information):**
 - **Transportr Version:** 2.1.1
 - **Device:** Android Emulator
 - **Android Version:** 4.1

grote:
maybe some constraintlayout regression/behavior change?

ialokim:
Already happens with Transportr Version 2.1.0 and seems to be specific to low API levels (at least >= API 19 it looks as expected).

ialokim:
There are also some other ugly UI issues like missing inset:
![Screenshot_1609083285](https://user-images.githubusercontent.com/18088991/103174333-824c5400-4861-11eb-8acc-350102e9dc4d.png)

I'd prefer to have a deeper look into this after the next release.

Altonss:
Is this bug still happenning?

ialokim:
I would guess so, yes. We didn't change anything in the UI layout. Anyway, supporting all those old Android versions is getting harder over time. Also taking into account that fewer and fewer people actually use those, we might focus on other things.

Altonss:
> I would guess so, yes. We didn't change anything in the UI layout. Anyway, supporting all those old Android versions is getting harder over time. Also taking into account that fewer and fewer people actually use those, we might focus on other things.

Any idea on which is the lowest API level we should target? I know F-Droid now has API 23 as minimum. I think under API level 21 is pretty old (< Android 5.0)... so maybe we could just drop support for these older versions?

grote:
feel free to bump minSdk to 21 (or higher if you want)

Issue #721, opened 2020-10-24
SSL certificate errors on API level 16-19


ialokim:
**Describe the bug**
On my local emulator with API level 16 (minSdk), Transportr refuses to search for locations and nearby stations with the DB provider.

**To Reproduce**
1. Select "DB" provider
2. Search for nearby stations from some point (e.g. in Berlin) on the map
3. Logcat reports the following:
```
10-24 15:56:15.520 4921-5031/de.grobox.liberario.debug I/de.grobox.transportr.locations.NearbyLocationsLoader: NearbyStation from (1000): Location{ADDRESS, 52.5198450/13.4073170, place=Berlin, name=Unnamed Road Unnamed Road}
10-24 15:56:15.760 4921-5031/de.grobox.liberario.debug W/System.err: javax.net.ssl.SSLHandshakeException: java.security.cert.CertPathValidatorException: Trust anchor for certification path not found.
10-24 15:56:15.770 4921-5031/de.grobox.liberario.debug W/System.err:     at org.apache.harmony.xnet.provider.jsse.OpenSSLSocketImpl.startHandshake(OpenSSLSocketImpl.java:374)
10-24 15:56:15.780 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.RealConnection.connectTls(RealConnection.java:320)
10-24 15:56:15.780 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.RealConnection.establishProtocol(RealConnection.java:284)
10-24 15:56:15.800 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.RealConnection.connect(RealConnection.java:169)
10-24 15:56:15.800 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.StreamAllocation.findConnection(StreamAllocation.java:258)
10-24 15:56:15.800 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(StreamAllocation.java:135)
10-24 15:56:15.800 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.StreamAllocation.newStream(StreamAllocation.java:114)
10-24 15:56:15.800 4921-5031/de.grobox.liberario.debug W/System.err:     at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.java:42)
[...]

Issue #471, opened 2018-07-21
Directions button hidden below bottom sheet on Android 4


Wikinaut:
On an old MEDION Lifetab S9512 with Android 4.0.4 (which I am currently testing), the navigation symbol is barely visible, and invisible, because covered, when having connections in the connections list.
![img_20180721_112139](https://user-images.githubusercontent.com/1151915/43034234-244972fc-8cd9-11e8-86b8-262e7790d92c.jpg)

* Android 4.0.4 (uh, yes...)
* Transportr 2.0.0 (installed from F-Droid on 20180721)

grote:
Thanks for reporting. I saw this as well on Android 4. It is probably related to `android:elevation` not working properly there. No time to debug exotic platform issues like those. Pull requests welcome!

PR #856, opened 2023-05-18
Bump minSdk to 21

Altonss:
This PR bumps minSdk to 21, and as such makes Transportr only compatible with Android devices running Android version 5.0 or above. This is done to simplify maintenance and fix bugs related to old versions.

This fixes:
- https://github.com/grote/Transportr/issues/721;
- https://github.com/grote/Transportr/issues/738;
- https://github.com/grote/Transportr/issues/471.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        api-level: [16, 23, 29]
+        api-level: [21, 23, 29]
 
     steps:
       - uses: actions/checkout@v2
--- a/app/src/debug/java/de/grobox/transportr/TransportrDebugApplication.kt
+++ b/app/src/debug/java/de/grobox/transportr/TransportrDebugApplication.kt
@@ -19,11 +19,9 @@
 package de.grobox.transportr
 
 import android.content.Context
-import androidx.multidex.MultiDex
 
 class TransportrDebugApplication : TransportrApplication() {
     override fun attachBaseContext(base: Context) {
         super.attachBaseContext(base)
-        MultiDex.install(this)
     }
 }
--- a/app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
+++ b/app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
@@ -36,22 +36,12 @@ class TransportrChangeLog(context: Context, settingsManager: SettingsManager) : 
     companion object {
 
         private fun theme(dark: Boolean): String {
-            return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
-                if (dark) {
-                    // holo dark
-                    "body { color: #e7e3e7; font-size: 0.9em; background-color: #292829; } h1 { font-size: 1.3em; } ul { padding-left: 2em; }"
-                } else {
-                    // holo light
-                    "body { color: #212421; font-size: 0.9em; background-color: #f7f7f7; } h1 { font-size: 1.3em; } ul { padding-left: 2em; }"
-                }
+            return if (dark) {
+                // material dark
+                "body { color: #f3f3f3; font-size: 0.9em; background-color: #424242; } h1 { font-size: 1.3em; } ul { padding-left: 2em; }"
             } else {
-                if (dark) {
-                    // material dark
-                    "body { color: #f3f3f3; font-size: 0.9em; background-color: #424242; } h1 { font-size: 1.3em; } ul { padding-left: 2em; }"
-                } else {
-                    // material light
-                    "body { color: #202020; font-size: 0.9em; background-color: transparent; } h1 { font-size: 1.3em; } ul { padding-left: 2em; }"
-                }
+                // material light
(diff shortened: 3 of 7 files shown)


Altonss:
@grote @pt2121 it looks good to me now, building fine :)

app/src/debug/java/de/grobox/transportr/TransportrDebugApplication.kt
```diff
@@ -19,11 +19,9 @@
 package de.grobox.transportr
 
 import android.content.Context
-import androidx.multidex.MultiDex
 
 class TransportrDebugApplication : TransportrApplication() {
     override fun attachBaseContext(base: Context) {
         super.attachBaseContext(base)
-        MultiDex.install(this)
```
ialokim, line 27: Great that we can get rid of this. Just leaving a pointer here to [the docs stating that the multidex support library is only necessary below API 21](https://developer.android.com/build/multidex).

app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
```diff
@@ -36,22 +36,12 @@ class TransportrChangeLog(context: Context, settingsManager: SettingsManager) :
     companion object {
 
         private fun theme(dark: Boolean): String {
-            return if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP) {
```
ialokim, line 39: Have you checked that this is the only point where we do special things for API < 21?

.github/workflows/build.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        api-level: [16, 23, 29]
+        api-level: [21, 23, 29]
```
ialokim, line 16: Maybe have one with the highest we support instead of 29? (Or is that still 29?)

ialokim (commented):
Looks good to me! Just some comments above.

app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
Altonss: I searched for "Build.VERSION" in the entire project, and couldn't find any other place... But if you know of any, I would be happy to add it!

app/src/debug/java/de/grobox/transportr/TransportrDebugApplication.kt
ialokim: No sorry, it was just meant to be here for future reference. (I've looked it up and wanted to document it somewhere.)

.github/workflows/build.yml
Altonss: Yes I though of the same thing, but I though we could maybe wait to do this with #853, and then do something like [21, 29, 33].

app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
ialokim: No, I don't. Thanks for the confirmation!

app/src/debug/java/de/grobox/transportr/TransportrDebugApplication.kt
Altonss: Perfect!

.github/workflows/build.yml
pt2121: My understanding is that these are the API levels of emulators that we want to run the tests against on GitHub workflow. We could update it

Altonss: [21, 29, 33] would be good to test the minSdk, something in between, and the max

ialokim: > So this PR can be merged as it is?

For me, yes.

> My understanding is that these are the API levels of emulators that we want to run the tests against on GitHub workflow. We could update it

Exactly. Maybe to the minimum API, the target API and some interesting API level in between where things could break. I don't know which one would be a good pick as the third one.

ialokim (approved):

.github/workflows/build.yml
pt2121: I'd do [21, 29, 31] then? since it's our target API

Altonss: > and some interesting API level in between where things could break.

Good idea, but I don't know which one :)

app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
Altonss: I checked again, and found a place that I overlooked. Just pushed a new commit to correct for it.

Altonss:
@ialokim somehow one test timed out, but I have not enough access to trigger a rerun, could you do it?

pt2121 (approved):
I didn't know I have a permission to approve a PR :slightly_smiling_face: 
Oh, well, not sure if it matters anyway.

Altonss:
> I didn't know I have a permission to approve a PR 🙂 ...not sure if it matters anyway.

Everyone can IIRC :)

grote:
Thanks everyone!

Issue #870, opened 2023-06-12
Find an alternative for lost


pt2121:
[lost](https://github.com/lostzen/lost) is deprecated. We should start looking for other options or solutions.
Let's find a library that is still maintained.

ialokim:
With #880 I actually don't use any third-party library anymore, but use Android's location information directly. Would be interesting to have more real-world tests if this is reliable (it has been for me).

ialokim:
Several users reported the GPS working well with #880, so I think we don't need to further consider a replacement for lost.

PR #880, opened 2023-08-23
Switch to maplibre v10 and jawg.io as map data source

ialokim:
This should enable us to get back into the main F-Droid repo eventually.

I would appreciate some testing by others before this is merged, especially concerning the device's location (GPS).


grote:
> I would appreciate some testing by others before this is merged, especially concerning the device's location (GPS).

If anybody gave this a test-run, please comment here.

rugk:
If anyone could upload a precompiled APK that would likely ease testing…

wallalex0:
I build a apk on commit a9d4ffa:
[transportr-debug.zip](https://github.com/grote/Transportr/files/12831752/transportr-debug.zip)

julianfairfax:
Thanks to @wallalex0's build, I can confirm that Transportr is fully functional with this PR, including the device's location over GPS (I do not have Google Play Services installed anyway, so this is all GNSS).

The biggest difference you notice between this version and the latest release, is that this version's map is fully greyscale. I'm not sure if this is in an intentional change or rather a technical limitation.

Personally I think the colour map is prettier, but a greyscale map is more useful, since it shows only the important details (i.e. the trip and stations), with other details fading into the background.

I know one test report won't seal the deal for you, but, if this were my project, I would definitely merge this as soon as possible, and get Transportr back in F-Droid as soon as possible.

It's a great app, with some unique features on Android (neither the DB or SBB apps have any map of the trip for example), and it's only going to get better. But, this PR, and the subsequent return to F-Droid, is a necessary first step.

(Maybe then @ialokim's SBB support can also be merged. Seems like https://github.com/schildbach/public-transport-enabler/pull/481 is the holdup. Once merged, Transportr will be the only open source app in existence to support SBB!)

rugk:
Also tested it and I can also agree:
* everything that worked before, seems to work, too
* also GPS works fine on the map at least (ref https://github.com/grote/Transportr/issues/870#issuecomment-1751119513), though have not tested outdoors; and the address did not resolve for me (though indoors it is hard and GPS generally not accurate where I am, so no problem here IMHO (old Transportr in the same place is different though, it resolves to any street in my case, which is the wrong one/place anyway, but well… this new Transportr-dev just shows "getting GPS")
* I also see the grayscale map, and I agree, IMHO a colored map similar to before would be more beautiful. It was already quite desaturated, so IMHO enough to see/focus on the actual route).

ialokim:
> The biggest difference you notice between this version and the latest release, is that this version's map is fully greyscale. I'm not sure if this is in an intentional change or rather a technical limitation.
> 
> Personally I think the colour map is prettier, but a greyscale map is more useful, since it shows only the important details (i.e. the trip and stations), with other details fading into the background.

That's a very good point, I had forgotten about this change (dark theme user here).

Here's how the difference looks in practice

Transportr-theme-old
Transportr-theme-new

There are more default map styles available from jawg.io, including `jawg-sunny` which matches the old style more closely: https://www.jawg.io/docs/apidocs/maps/dynamic-maps/#jawg-default-styles

On the other hand, the used `jawg-light` theme corresponds better to the dark map style that is used in the dark theme version of the app, which is also grayscale.

julianfairfax:
> There are more default map styles available from jawg.io, including jawg-sunny which matches the old style more closely: https://www.jawg.io/docs/apidocs/maps/dynamic-maps/#jawg-default-styles

> On the other hand, the used jawg-light theme corresponds better to the dark map style that is used in the dark theme version of the app, which is also grayscale.

I would still switch to `jawg-sunnay` since it's consistent with the old version, and it looks prettier.

1Maxnet1:
> I build a apk on commit [a9d4ffa](https://github.com/grote/Transportr/commit/a9d4ffaac9b80a9538da109a1bfb81477a3dad7b): [transportr-debug.zip](https://github.com/grote/Transportr/files/12831752/transportr-debug.zip)

I also gave this a test run in my daily travels the last days and everything seems to work as expected. Thanks for the work!

bernhardreiter:
@grote  @ialokim  now that we had three test reports, all positive, are you confident enough to merge ?

ialokim:
I have changed the light theme now to `jawg-sunny` following the suggestions from above. For reference, it now looks like this:

Transportr-theme-new2

I will proceed to merging when the CI build succeeds.

Issue #826, opened 2022-08-15
Update fdroid release


leg7:
The last fdroid release was 4 years ago and even using the custom repository in the readme doesn't fix this.

Can the fdroid version be updated?

julianfairfax:
This is the supposed solution: https://gitlab.com/fdroid/fdroidserver/-/issues/1004#note_1036153553

leg7:
So from what I got a solution is being worked on to replace the google maps dependency with maplibre?

Is google maps a proprietary blob included in the apk or is transportr just using the api? Would it be feasible to get rid of the proprietary dependency upstream?

julianfairfax:
> So from what I got a solution is being worked on to replace the google maps dependency with maplibre?
> 

I don't know if it's being worked on but it has been suggested.
> Is google maps a proprietary blob included in the apk or is transportr just using the api? Would it be feasible to get rid of the proprietary dependency upstream?

Mapbox is dragging it in. I suppose it would be possible to fix that upstream, but if switching to MapLibre works, that's probably a better solution.

ialokim:
Yes, we're aiming to switching to MapLibre since new versions of Mapbox itself ar not compatible with free software anymore. Still default MapLibre currently has the same problem of depending on proprietary parts at compile time, which is not accepted by F-Droid. See the upstream issue at https://github.com/maplibre/maplibre-gl-native/issues/379 and #668 for our part, switching to MapLibre.

jakoblm:
The issue has now been pushed again in the MapLibre repo: https://github.com/maplibre/maplibre-gl-native/issues/379

jakoblm:
How about Transportr switching to the alternative AnyMaps (https://github.com/johan12345/AnyMaps) or osmdroid (https://github.com/osmdroid/osmdroid)? Obviously both use only free software and are accepted by F-Droid without any problems. See the following comment: https://github.com/maplibre/maplibre-gl-native/issues/379#issuecomment-1250773381

1Maxnet1:
A short time workaround would be to offer stable releases over a third party repository (e.g. https://android.izzysoft.de/repo/info offers to include apps in his repository). This would enable users to add the repository and use a more modern app version.

ialokim:
[...]

Issue #768, opened 2021-07-24
Switzerland: SBB Not work.


Mannshoch:
**Describe the bug**
If I use SBB as provider Transportr do not react on entry any citys.

**To Reproduce**
1. Go to Routeplaning 
2. Enter a city
3. Nothing happend

**Expected behavior**
Dropdown wit trainstations in Zürich appearr.

**Versions (please complete the following information):**
 - **Transportr Version:** 2.1.2
 - **Device:** Samsung S7
 - **Android Version:** Lineageos 18

SBB API: 
https://data.sbb.ch/api/v2/console
https://data.sbb.ch/api/v1/console

ialokim:
Thanks for your report, I was able to reproduce the problem. Most probably, SBB changed something on their end so that [PTE](https://github.com/schildbach/public-transport-enabler) (the library Transportr uses under the hodd) is unable to access the information. You might consider opening a bug report over there as well.

sirtoobii:
Hej @Mannshoch  I've implemented [timetable.search.ch](timetable.search.ch) as a SBB replacement. Are you willing to participate in a "closed beta"-test? If yes, please message me at tobias[dot]bossert[at]fastpath.ch[dot].ch, I will then send you an apk of Transportr with the new Provider included

sirtoobii:
Hej testers! Thank you for your feedback. A new release is available with the following fixes:
- We recognize more transportation types now
- The nearby stations feature should work now
- The stationboard-feature should work for every station now
- No more "null" lines

You can download the new beta (**unofficial!**) release here: https://sun.fastpath.ch/index.php/s/PLs3xc8j7nzGm4i

iGormilhit:
@sirtoobii I tried your unofficial beta, and it works like a charm. Thanks.

grote:
Note that PTE upstream has SBB removed. If you want it back in Transportr, one would need to reach out to PTE for the reasons and depending on the answer add it back to the PTE staging fork Transportr is using.

sirtoobii:
The reason was the shutdown of SBB's hafas endpoint (https://github.com/schildbach/public-transport-enabler/issues/381). A merge of https://github.com/schildbach/public-transport-enabler/pull/383 would resolve the issue.

grote:
Cool, then you could consider submitting it to https://gitlab.com/opentransitmap/public-transport-enabler/ to get it into Transportr quicker.

sirtoobii:
[...]

PR #689, opened 2020-08-04
Release 2.1.0

ialokim:
As requested in #683, I've prepared the next release.

I've updated PTE and the translations, bumped the version number and tested the release.

As some Guava annotations are used directly in Transportr source code, I've added it now as an explicit dependency directly excluding the submodules that are not needed.

Fixes #683

PR #872, opened 2023-06-12
Material Design 3 migration

Bnyro:
Hey,

I recently saw this project has been revived and have been really happy to see the development continue.
Since the app's current design is already some years behind the current trend, I thought it'd be awesome to start with a new, improved and more modern UI.

Therefore this PR attempts to move the app to [Material Design 3](https://m3.material.io). Please note that I haven't done any Unit or UI tests and hence can't tell if some of them are broken now (in theory there shouldn't be anything in these changes causing it!).

Please note that the changes required me to update some dependencies (AGP to 7.4.2, Material library to 1.7.0, Gradle to 7.5, appcompat to 1.5.1) and the SDK to 32. I hope there aren't any new issues coming up with that, at least from my testing it seems fine.

There are still two issues I'm fighting with and can't really figure out how to fix because I'm not that familiar with the app yet:
1. When changing the network in the navigation drawer, the UI freezes and the app crashes after some seconds. However, after restarting the app, the wanted network got selected. That's been related to the dependency updates too, fixed by downgrading some libs.
2. Now tracked in #873

I hope that changes are welcome (since I forgot to open an issue first) and that someone can test the changes or assist with the two bugs above. Of course I don't claim the changes to be perfect, but in my eyes the app looks pretty neat after applying these.

Cheers 

Edit: The screenshots below were made on Android 13, so the app uses dynamic wallpaper based colors. For Android 11 and below, the colors should be similar to the current ones as they're based on the former primary color and generated with https://m3.material.io/theme-builder.

PR #888, opened 2023-11-12
Release 2.2.0

ialokim:
Here are the steps to follow when preparing a new release (some custom tasks added for this release). Please check the following boxes with an `x` when done:

* [x] update PTE to the latest commit at [opentransitmap](https://gitlab.com/opentransitmap/public-transport-enabler/-/commits/master) and run `./update-dependency-pinning.sh`
* [x] re-enable SBB (Switzerland) with new data source #768 
* [x] ~~re-enable PL (Poland) with new data source https://gitlab.com/opentransitmap/public-transport-enabler/-/commit/693d84e340f1525366e6b29a32c0a81401f326ff~~ (needs more investigation as to which area is actually covered)
* [x] test the app (automatically using AndroidTest as well as manually), paying special attention to changes since the last release
* [x] fix any bugs if necessary
* [x] update translations from [Transifex](https://www.transifex.com/) using `tx pull --mode=developer -a` in the root folder (you need proper permissions to do that)
* [x] add newly supported languages to the two arrays in [`app/src/main/res/values/arrays.xml`](https://github.com/grote/Transportr/blob/master/app/src/main/res/values/arrays.xml#L16)
* [x] revise the commits to `master` since the last release and add interesting changes (as well as new languages) to the changelog at [`app/src/main/res/xml/changelog_master.xml`](https://github.com/grote/Transportr/blob/master/app/src/main/res/xml/changelog_master.xml), then run `python3 ./fastlane/generate_changelog.py`
* [x] bump the [`versionCode`](https://github.com/grote/Transportr/blob/master/app/build.gradle#L14) and [`versionName`](https://github.com/grote/Transportr/blob/master/app/build.gradle#L15) in `app/build.gradle`
* [ ] re-enable Transportr for inclusion in F-Droid #826
* [ ] file another PR updating all dependencies which should be merged *after* the successful release

Please also refer to #689 as an example.


ialokim:
> * [x] fix any bugs if necessary

several issues reported: #889 #890 https://gitlab.com/opentransitmap/public-transport-enabler/-/issues/6

In order to get the release out soon, I'd consider them as been non-release-blocking.

M00NJ:
Could #872 be a part of this release as well?

ialokim:
> Could #872 be a part of this release as well?

I don't think this is a good idea, given that we want to publish the release soon, and #872 brings a lot of UI changes that need a thourough testing.

mesinger:
@ialokim any way i can help to get this release done?

app/src/main/res/xml/changelog_master.xml
```diff
@@ -2,6 +2,17 @@
 <changelog xmlns:tools="http://schemas.android.com/tools"
     tools:ignore="UnusedResources">
 
+    <release
+        version="2.2.0"
+        versioncode="124">
+        <change>Switch map library to MapLibre - finally bringing Transportr back to F-Droid!</change>
```
grote, line 8: Technically, we didn't make it back, yet ;)

grote (approved):

app/src/androidTest/java/de/grobox/transportr/ScreengrabTest.kt
```diff
@@ -51,17 +51,13 @@ abstract class ScreengrabTest {
         Screengrab.setDefaultScreenshotStrategy(UiAutomatorScreenshotStrategy())
     }
 
-    val networkId: NetworkId = when(Locale.forLanguageTag(getTestLocale())) {
+    val networkId: NetworkId = when(Locale.forLanguageTag(getTestLocale() ?: "de")) {
```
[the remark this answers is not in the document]
ialokim: I can't fully remember right now why this was, but I think `getTestLocale()` would return `null` in certain cases and therefore fail the test. Anyway it's just in the test sources which require some overhaul after removing quite some regions, so maybe not a blocker either?

ialokim:
Thanks for your review @Altonss!

>     * [ ]  Why updating strings for pl network navitia that is for now removed? IMHO it clutters a bit the PR with useless strings.

There are a lot of leftover strings from now removed regions which should probably be cleaned properly at some point. But you are right, those are unfortunately really useless. I've removed them now manually.

>     * [ ]  Shouldn't we also mention in the changelogs that we've re-added VRS?

Good catch, added that!

app/src/main/res/xml/changelog_master.xml
ialokim: Yeah it's the next unchecked bullet point in the first post, but we cannot continue without tagging the release :P

ialokim: And we will make it back (I hope)! :smile:

grote:
Thanks everyone who contributed to this release. You are awesome! :D

I just pressed the button and the release is out. However, Google Play doesn't accept it: `Google Api Error: Invalid request - Target SDK of artifact is too low: 124.` It seems it is currently 32 which won't work anymore. Please see https://support.google.com/googleplay/android-developer/answer/11926878 for details.

Issue #890, opened 2023-11-14
TimePicker minutes are always set to current time


ialokim:
**Describe the bug**
The minute part of the time picker in the trip search always defaults to the current time instead of the previously selected one.

**To Reproduce**
1. search for any connection
2. change the departure/arrival time, especially the minute part
3. try to change the time again, notice that the minute part is automatically adjusted to the current time

**Expected behavior**
Both the hour and minute part of the time picker should coincide with the previously selected time.

**Versions (please complete the following information):**
 - **Transportr Version:** 2.2.0 (preview)
 - **Device:** Google Pixel 4a
 - **Android Version:** 13

PR #893, opened 2023-12-06
Fix timePicker minutes always set to current time

Altonss:
This PR fixes issue #890.

### DIFF
--- a/app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
@@ -21,6 +21,7 @@ package de.grobox.transportr.ui
 
 import android.app.DatePickerDialog
 import android.app.DatePickerDialog.OnDateSetListener
+import android.os.Build
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


ialokim:
Thanks for the fix! Could you elaborate here (or even better add a comment) as to why this fixes the issue? Maybe a link to the timePicker Android docs?

Please also squash your commits into one for them to be merged.

ialokim:
Hum sorry I still don't get it. Is it a known issue with the deprecated `setCurrent{Hour,Minute}()` methods on API higher than 23, or what variable was reset when changing either hour or minute?

ialokim:
>     1. The main issue: the c calendar variable gets reset when setting either hour or minute for timePicker. I fixed it by saving the values of the c variable before setting timePicker.

Ah, now I got it - I looked at the surrounding code and found the `TimeChangedListener` that updates the calendar. Nice catch!

>     2. the deprecated API used for setCurrent which isn't really an issue, I just cleaned the code for the future.

Thanks for that too.

ialokim (approved):

PR #894, opened 2023-12-06
Fix potential UI glitch when transport type is not given

### LINKED ISSUE #827: Potential UI glitch when transport type is not given
see https://gitlab.com/opentransitmap/public-transport-enabler/-/merge_requests/10#note_1053067737

image

Altonss:
Fixes #827

### DIFF
--- a/app/src/main/java/de/grobox/transportr/utils/TransportrUtils.kt
+++ b/app/src/main/java/de/grobox/transportr/utils/TransportrUtils.kt
@@ -53,7 +53,7 @@ object TransportrUtils {
         FERRY -> R.drawable.product_ferry
         CABLECAR -> R.drawable.product_cablecar
         ON_DEMAND -> R.drawable.product_on_demand
-        null -> R.drawable.product_bus
+        null -> R.drawable.product_null
     }
 
     @JvmStatic


app/src/main/res/drawable/product_null.xml
```diff
@@ -0,0 +1,8 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+        android:width="3.5dp"
+        android:height="24dp"
+        android:viewportWidth="3.5"
+        android:viewportHeight="24.0">
+    <path
+        android:fillColor="#FF000000"/>
```
ialokim, line 7: How does the icon look like for a missing transport type? Just a white box? Or what does this do?

ialokim (commented):
Thanks for this fix!

app/src/main/res/drawable/product_null.xml
ialokim: Thanks for posting a screenshot, looks good indeed. But I'm still not fully convinced by the empty icon, feels a bit like a hack. I think the cleaner solution would be to change the layout file that defines the product view to ensure it works even without an icon.

I'd suggest leaving this for post-2.2.0.

ialokim:
> Thanks for posting a screenshot, looks good indeed. But I'm still not fully convinced by the empty icon, feels a bit like a hack. I think the cleaner solution would be to change the layout file that defines the product view to ensure it works even without an icon.

@Altonss do you feel like looking into the layout file again?

PR #895, opened 2023-12-06
Hide locations swap icon when fields are empty

Altonss:
Solves #732


app/src/main/java/de/grobox/transportr/trips/search/DirectionsFragment.kt
```diff
@@ -111,8 +111,10 @@ class DirectionsFragment : TransportrFragment() {
             fromLocation.setLocation(it)
             if (it != null) toLocation.requestFocus()
         })
+        viewModel.fromLocation.observe(viewLifecycleOwner, { onLocationsChanged() })
         viewModel.viaLocation.observe(viewLifecycleOwner, { viaLocation.setLocation(it) })
         viewModel.toLocation.observe(viewLifecycleOwner, { toLocation.setLocation(it) })
+        viewModel.toLocation.observe(viewLifecycleOwner, { onLocationsChanged() })
```
ialokim, line 117:
```suggestion
        viewModel.toLocation.observe(viewLifecycleOwner, {
            toLocation.setLocation(it)
            onLocationsChanged()
        })
```

Same suggestion for above.

ialokim (changes requested):
Thanks for this!

As discussed in https://github.com/grote/Transportr/pull/733#issuecomment-748000401 I'm still not convinced that removing the icon button completely is the right thing to do. Especially because there is another icon left of it that might jump around.

In any case I would postpone this fix until after the release.

ialokim:
> > As discussed in [#733 (comment)](https://github.com/grote/Transportr/pull/733#issuecomment-748000401) I'm still not convinced that removing the icon button completely is the right thing to do. Especially because there is another icon left of it that might jump around.
> > In any case I would postpone this fix until after the release.
> 
> Oh yes this would make sense, so preferably just disable this button right?

Yes, but "just" disabling might not be that easy, in the last PR we didn't find a proper way of visually disabling an IconButton, iirc.

ialokim:
I'm mostly concerned about the favorite icon jumping around if this gets hidden. If swap is only ever hidden when favorite is hidden, too, I think I could live with it.

PR #904, opened 2024-01-01
Reenable Wiener Linien provider

mesinger:
Reenables WienerLinien provider as it works on version 2.1.5 without any problems


ialokim (approved):
LGTM too! Tested and indeed works without problems. Thanks @mesinger !

PR #910, opened 2024-01-10
Update target SDK to 33

Altonss:
No significant changes affecting Transportr: https://developer.android.com/about/versions/13/behavior-changes-13


grote:
Thanks for the quick PR! Should still get tested thoroughly, because easy to miss SDK changes that require action.

grote (approved):

ialokim:
I've successfully tested this on Android 14 and didn't find any regressions.

PR #914, opened 2024-01-11
Fix some theming-related issues

### LINKED ISSUE #909: Background color of changelog is dark when in light mode.
Background color of changelog is dark when in light mode.
Noticed here: https://github.com/grote/Transportr/pull/888#pullrequestreview-1807893223

 - **Transportr Version:** 2.1.5 & 2.2.0
 - **Android Version:** 13

### LINKED ISSUE #913: Default theme setting is ignored
**Describe the bug**
After the initial setup of the app, the used app theme differs from the theme value set in the settings by default.

**To Reproduce**
1. Set your system theme to dark mode (the system theme should not equal the default light theme setting)
2. (optionally) Clean the app data
3. Open the app and select any transportation network
4. Go to the app settings and look at the theme setting

**Expected behavior**
The initial theme setting seems to be "Light", so the app should be in light mode.

**Actual behavior**
The app uses the system theme instead. If you switch the system to the light mode and back to the dark mode again, the default theme setting is respected by the app.

 - **Transportr Version:** 2.2.0
 - **Device:** Google Pixel 6a
 - **Android Version:** GrapheneOS (based on the latest AOSP release of Android 14)

ialokim:
- fixes #913 by changing the default to "auto" (follow system)
- fixes #909 by only checking for system-wide theme if "auto" is enabled

### DIFF
--- a/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
+++ b/app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
@@ -60,9 +60,9 @@ class SettingsFragment : PreferenceFragmentCompat() {
 
         // Fill in current transport network if available
         networkPref = findPreference("pref_key_network")!!
-        manager.transportNetwork.observe(this, Observer<TransportNetwork> {
+        manager.transportNetwork.observe(this) {
             onTransportNetworkChanged(it)
-        })
+        }
 
         networkPref.setOnPreferenceClickListener {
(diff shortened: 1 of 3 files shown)


ialokim:
Just tested locally on a Android 5 (API 21) powered emulator. Fixed a bug where dark mode would not be recognized for Android <10 when battery saver mode is enabled. Everything else worked as expected.

PR #923, opened 2024-01-20
Poc/draw vienna subway

mesinger:
image

Directly seeing the lines on the map improves a user's orientantion when they open the app.
Many cities color-code their lines so this is also an important information that is currently completely missing in Transportr.
As this feature is available in mainstream Google/Apple Maps i think this is an important feature that would improve the usability of Transportr.

The screenshot shows the 5 subway lines in Vienna in their correct color, so anyone living in Vienna will directly know which color is which line.

DISCLAIMER: The coordinates for the lines are parsed from the Overpass API. As I'm new to the development of map applications their most likely is a better solution to query the coordinates of transportation lines. This is just a POC.


mesinger:
ℹ️ I did that poc on my kotlin1.8 PR so the changes related to this feature are in this [commit](https://github.com/grote/Transportr/pull/923/commits/220363a87c3b0cda79dd8b4ef5cd42f0b1e45817)

mesinger:
@ialokim @Altonss before i invest more time in this i need to get some feedback from you.

ialokim:
I do like the idea of showing more public-transport-related information readily on the map, but I don't think manually querying data from OSM is a way that would scale to other regions. It also leads to questions as for which regions special information is shown, why limiting this to metro and not including buses etc.

I was hoping that now that we switched to jawg.io as our map provider with #668 , we could try to customize the map styles we are using to highlight public transport information. However, I just had a quick look at their style editor and it seems very limited in what we can customize there.

On a larger roadmap, we could look into https://protomaps.com/ and if it would be possible to highlight custom information with that.

PR #927, opened 2024-01-31
Material Design 3 migration (V2)

newhinton:
Closes #872 

This PR is based on, and superseeds @Bnyro 's Material 3 implementation. It takes the UI a step further, and makes heavy use of the Material You colors.

This was implemented and improved with @Bnyro support, consent and collaboration.

Edit:
For more info about what this pr does, look in this discussion: https://github.com/Bnyro/Transportr/pull/1


newhinton:
I agree to the CLA

newhinton:
Todo:

In the trip detail view, warnings/notes are no longer colored in the warning color

michaelblyons:
Spelled your name wrong in the JSON file.

.clabot
```diff
@@ -22,7 +22,8 @@
     "thgoebel",
     "pt2121",
     "Jean-BaptisteC",
-    "Bnyro"
+    "Bnyro",
+    "newinton"
```
michaelblyons, line 26: Missing an "h" ?

newhinton:
aa04e25 adds better transitions to the details view, which should improve UX by quite a bit.


If it would work properly. 
Currently the view is jumping around quite a bit (by switching states from middle back to full) but i dont know why. Generally speaking the transitions are a lot nicer than before. Any help is appreciated!

newhinton:
Known issues:

- Select Location from main searchbar, dismiss bottom sheet. Re-select location. Left side of bottom sheet has padding
- Drop shadow of navigation also has padding after above flow
- Navigation Drawer is invisible below status bar when opened
- Three State Bottom Sheet has buggy behaviour
- Select Location on map. Start navigation via FAB. New Activity has favorite locations instead of empty list (already present in 2.2.1)

ialokim:
Thanks @newhinton and @Bnyro! The screenshots in the linked PR already look very promising!

I will try to test your changes in the coming days and provide some initial feedback.

app/src/main/java/de/grobox/transportr/map/MapActivity.java
```diff
@@ -74,14 +90,23 @@ public class MapActivity extends DrawerActivity implements LocationViewListener
...
+		setupDrawer();
+		FullScreenUtil.Companion.drawBehindStatusbar(this);
+		FullScreenUtil.Companion.applyTopInset(findViewById(R.id.searchCardView), 16);
 
 		View menu = findViewById(R.id.menu);
-		menu.setOnClickListener(view -> openDrawer());
+		menu.setOnClickListener(view -> {
+			InputMethodManager imm = (InputMethodManager) getSystemService(Activity.INPUT_METHOD_SERVICE);
+			View currentFocus = getCurrentFocus();
+			if (currentFocus != null) imm.hideSoftInputFromWindow(currentFocus.getWindowToken(), 0);
```
ialokim, line 106: As before, I would move this to `KeyboardUtils`.

app/src/main/java/de/grobox/transportr/map/MapActivity.java
```diff
@@ -74,14 +90,23 @@ public class MapActivity extends DrawerActivity implements LocationViewListener
...
 	public void onCreate(@Nullable Bundle savedInstanceState) {
+		SplashScreen.installSplashScreen(this);
 		super.onCreate(savedInstanceState);
 		if (BuildConfig.DEBUG) enableStrictMode();
 		getComponent().inject(this);
 		setContentView(R.layout.activity_map);
-		setupDrawer(savedInstanceState);
+		setupDrawer();
+		FullScreenUtil.Companion.drawBehindStatusbar(this);
+		FullScreenUtil.Companion.applyTopInset(findViewById(R.id.searchCardView), 16);
```
ialokim, line 100: This misbehaves on my device, see screenshots.

app/src/main/java/de/grobox/transportr/networks/PickTransportNetworkActivity.kt
```diff
@@ -61,7 +63,7 @@ class PickTransportNetworkActivity : TransportrActivity(), ISelectionListener<II
         } else {
             firstStart = false
             supportActionBar?.setDisplayHomeAsUpEnabled(true)
-            findViewById<View>(R.id.firstRunTextView).visibility = GONE
+            findViewById<TextView>(R.id.firstRunTextView).text = ""
```
ialokim, line 66: Why did you change this? This results in a way too large height of the header.

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
```diff
@@ -108,4 +110,24 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
     }
 
+    override fun onDisplayPreferenceDialog(preference: Preference?) {
+        when (preference) {
+            // show a material design 3 dialog instead of the default md2 preference dialog
```
ialokim, line 115: The rest of preferences looks pretty much like MD2 still.

app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
```diff
@@ -194,9 +195,10 @@ internal class LegViewHolder(v: View, private val listener: LegClickListener, pr
         lineView.setWalk()
 
         // line color
-        fromCircle.setColorFilter(getColor(fromCircle.context, R.color.walking))
-        lineBar.setColorFilter(getColor(lineBar.context, R.color.walking))
-        toCircle.setColorFilter(getColor(toCircle.context, R.color.walking))
+        val color = MaterialColors.getColor(fromCircle.context, R.attr.colorSecondary, Color.TRANSPARENT)
+        fromCircle.setColorFilter(color)
+        lineBar.setColorFilter(color)
+        toCircle.setColorFilter(color)
```
ialokim, line 201: Not sure if I like having the walking color follow the theme.

app/src/main/java/de/grobox/transportr/ui/TimeDateFragment.kt
```diff
@@ -62,6 +61,11 @@ class TimeDateFragment : DialogFragment(), OnDateSetListener, OnTimeChangedListe
         }
     }
 
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        return MaterialAlertDialogBuilder(requireContext())
+            .show()
+    }
```
ialokim, line 67: Couldn't `DialogFragment` use `MaterialAlertDialogBuilder` by default?

app/src/main/java/de/grobox/transportr/ui/TransportrChangeLog.kt
```diff
@@ -20,28 +20,58 @@
...
 
 class TransportrChangeLog(context: Context, settingsManager: SettingsManager) : ChangeLog(
-    ContextThemeWrapper(context, R.style.DialogTheme),
+    context,
     settingsManager.settings,
     theme(settingsManager.isDarkTheme)
 ) {
+    fun getMaterialDialog(full: Boolean): AlertDialog {
+        val wv = WebView(mContext)
+        wv.setBackgroundColor(Color.TRANSPARENT)
```
ialokim, line 40: Nice fix!

app/src/main/java/de/grobox/transportr/utils/DateUtils.kt
```diff
@@ -71,7 +73,7 @@ object DateUtils {
         val delayMinutes = millisToMinutes(delay)
         return Delay(
             delay = "${if (delayMinutes >= 0) '+' else ""}$delayMinutes",
-            color = ContextCompat.getColor(context, if (delayMinutes > 0) R.color.md_red_500 else R.color.md_green_500)
+            color = MaterialColors.getColor(context, if (delayMinutes > 0) R.attr.colorError else R.attr.colorPrimary, Color.TRANSPARENT)
```
ialokim, line 76: Not sure if the Material You colors are meant to be used as text color?

app/src/main/java/de/grobox/transportr/utils/FullScreenUtil.kt
```diff
@@ -0,0 +1,107 @@
...
+import androidx.core.view.WindowCompat
+import androidx.core.view.WindowInsetsCompat
+import de.grobox.transportr.utils.TransportrUtils.dpToPx
+
+
+class FullScreenUtil {
+    companion object {
+        fun drawBehindStatusbar(activity: Activity) {
+            WindowCompat.setDecorFitsSystemWindows(activity.window, false)
+            // required for the mail view, which cant properly use the theme.
```
ialokim, line 42: which mail view?

app/src/main/java/de/grobox/transportr/utils/FullScreenUtil.kt
```diff
@@ -0,0 +1,107 @@
...
+                    insets.getInsets(WindowInsetsCompat.Type.statusBars()).top
+                } else {
+                    insets.systemWindowInsetTop
+                }
+
+                val parameters = CoordinatorLayout.LayoutParams(
+                    v.layoutParams.width,
+                    v.layoutParams.height
+                )
+                parameters.setMargins(dpToPx(view.context, dpLeft), top + originalOffset, dpToPx(view.context, dpRight), dpToPx(view.context, dpBottom))
```
ialokim, line 82: I would suspect this rather black magic to cause the problems with the trip details view. Where did you get this from? Shouldn't there be an easier way for achieving this?

And why wasn't it necessary before?

app/src/main/res/drawable/ic_action_departures.xml
```diff
@@ -1,6 +1,7 @@
 <vector xmlns:android="http://schemas.android.com/apk/res/android"
         android:width="24dp"
         android:height="24dp"
+        android:tint="?colorControlNormal"
```
ialokim, line 4: All those icons are invisible now on my device.

app/src/main/res/drawable/ic_launcher_foreground_primary_color.xml
```diff
@@ -0,0 +1,14 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
ialokim, line 1: Shouldn't it be possible to reuse the existing `ic_launcher_foreground.xml` and just apply a custom tint to it wherever this is needed?

app/src/main/res/drawable/product_bus.xml
```diff
@@ -2,7 +2,8 @@
         android:width="24dp"
         android:height="24dp"
         android:viewportWidth="24.0"
-        android:viewportHeight="24.0">
+        android:viewportHeight="24.0"
+        android:tint="?attr/colorOnBackground">
```
ialokim, line 6: Was there a difference between `?attr/color*` and `?color*`? if not, might be worth being consistent.

app/src/main/res/drawable/shape_location_dropdown.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+	android:shape="rectangle">
+	<solid android:color="@color/md_theme_background" />
+	<stroke
+		android:width="1dp"
```
ialokim, line 6: worth moving this to `dimen`?

app/src/main/res/layout/activity_directions.xml
```diff
@@ -1,40 +1,48 @@
 <?xml version="1.0" encoding="utf-8"?>
-<androidx.coordinatorlayout.widget.CoordinatorLayout
-	xmlns:android="http://schemas.android.com/apk/res/android"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ialokim, line 2: I hope you know what you're doing :wink:  this might affect the whole bottomSheetBehavior too.

app/src/main/res/layout/activity_map.xml
```diff
@@ -1,76 +1,89 @@
...
+		android:id="@+id/navigationView"
+		android:layout_width="wrap_content"
+		android:layout_height="match_parent"
+		android:layout_gravity="start"
+		android:fitsSystemWindows="false"
+		app:headerLayout="@layout/header_navigation_drawer"
+		app:menu="@menu/drawer_menu" />
 
-</androidx.coordinatorlayout.widget.CoordinatorLayout>
+</androidx.drawerlayout.widget.DrawerLayout>
```
ialokim, line 89: Could you change your IDE settings to force a trailing newline?

app/src/main/res/layout/fragment_directions_form.xml
```diff
@@ -150,7 +147,6 @@
 		android:paddingEnd="4dp"
 		android:paddingRight="4dp"
 		android:gravity="center"
-		android:textColor="@color/md_grey_300"
```
ialokim, line 153: I don't think it's a good idea to have text color set by Material You.

app/src/main/res/layout/header_navigation_drawer.xml
```diff
@@ -0,0 +1,70 @@
...
+<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto"
+	style="@style/NavHeaderCard"
+	android:layout_width="match_parent"
+	android:layout_height="match_parent"
+	android:layout_margin="8dp"
+	android:layout_marginTop="32dp"
+	android:fitsSystemWindows="true">
+
+	<LinearLayout
```
ialokim, line 11: Couldn't we use a `ConstrainedLayout` instead of nested `LinearLayout`s?

app/src/main/res/layout/list_item_favorite.xml
```diff
@@ -45,47 +44,47 @@
...
+			tools:visibility="visible"
+			style="@style/FavoriteCardForeground" />
 
 		<TextView
 			android:id="@+id/via"
 			android:layout_width="0dp"
 			android:layout_height="wrap_content"
 			android:ellipsize="middle"
 			android:singleLine="true"
-			android:textColor="?android:textColorPrimary"
```
ialokim, line 73: As said before, not a fan of having Material You-adapted text colors.

app/src/main/res/values-v31/colors.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+
+    <color name="md_theme_primary">@android:color/system_accent1_200</color>
+    <color name="md_theme_onPrimary">@android:color/system_accent1_0</color>
```
ialokim, line 5: Where did you get these values from? Some online tool?

app/src/main/res/values/themes.xml
```diff
@@ -1,62 +1,55 @@
...
         <item name="windowActionModeOverlay">true</item>
-        <item name="android:dialogTheme">@style/DialogTheme</item>
-        <item name="android:alertDialogTheme">@style/DialogTheme</item>
-        <item name="popupTheme">@style/ThemeOverlay.AppCompat.DayNight</item>
+        <item name="android:statusBarColor">?android:attr/colorBackground</item>
+        <item name="android:navigationBarColor">?android:attr/colorBackground</item>
     </style>
 
     <style name="AppDayNightTheme" parent="AppBaseTheme">
-        <item name="colorControlNormal">@color/drawableTintLight</item>
```
ialokim, line 20: Shouldn't this still be set somewhere or is it provided by MD3 somehow automatically? (You've used it in several places)

app/src/test/java/de/grobox/transportr/utils/DateUtilsTest.kt
```diff
@@ -79,8 +79,6 @@ class DateUtilsTest {
     fun initMocks() {
         MockitoAnnotations.initMocks(this)
         `when`(context.resources).thenReturn(resources)
-        `when`(resources.getColor(R.color.md_green_500)).thenReturn(GREEN)
-        `when`(resources.getColor(R.color.md_red_500)).thenReturn(RED)
```
ialokim, line 83: Have you tried that the test still runs through?

ialokim (changes requested):
Hi @newhinton and @Bnyro, I've now gone through the code and left some comments below. Thanks again for this work!

ialokim:
>  - Select Location from main searchbar, dismiss bottom sheet. Re-select location. Left side of bottom sheet has padding

I haven't encountered that one.

> - Drop shadow of navigation also has padding after above flow
> - Navigation Drawer is invisible below status bar when opened
> - Three State Bottom Sheet has buggy behaviour

But those ones yes.

> - Select Location on map. Start navigation via FAB. New Activity has favorite locations instead of empty list (already present in 2.2.1)

Given this is not a regression, I wouldn't further consider it for this PR.

> - ~Trip Details toolbar menu does not show up~

This one still happens here.

ialokim:
Additionally, I've found the following issues while testing:

- Text color looks strange when selected via Material You:


- Issues with transport mode selection: not MD3 dialog, icon alignment


- DatePicker dialog has different shape


- map attribution is partly covered by "home screen bar"


- I don't think the back button needs to be as prominent, I'd just leave it in the onSurface color:


- text color issue on network selection page, also unnecessary high top bar


- invisible icons in several places of the app

ialokim:
This definitely needs some more cleanup work, but I do think this is going into the right direction!

Personally, I'm not fully convinced on giving up on the nice Transportr red altogether, given that it's part of the branding, but it seems that's what Material You is suggesting to do?

app/src/main/java/de/grobox/transportr/settings/SettingsFragment.kt
Bnyro: Yep, Material Design 3 unfortunately doesn't support preference styling, that's why we need to do that dialog workaround. If we wanted to really make the preferences MD3, we would need to write the preference fragments from scratch, which probably nobody has interest in doing...

app/src/main/java/de/grobox/transportr/utils/DateUtils.kt
Bnyro: Yes, see "Primary" roles at https://m3.material.io/styles/color/roles

app/src/main/res/drawable/product_bus.xml
Bnyro: It's the same, we should only use one way for consistency.

app/src/main/res/values-v31/colors.xml
Bnyro: https://m3.material.io/theme-builder

app/src/main/res/values/themes.xml
Bnyro: It's automatically set by Android

Bnyro:
> Personally, I'm not fully convinced on giving up on the nice Transportr red altogether, given that it's part of the branding, but it seems that's what Material You is suggesting to do?
>
Yes, it does. It would be possible to add s preference to toggle the red vs material you colors though.

michaelblyons:
> It would be possible to add a preference to toggle the red vs Material You colors though.

I think this is what Trail Sense is doing.

app/src/main/java/de/grobox/transportr/trips/detail/LegViewHolder.kt
newhinton: You are right, it leads to some readability issues. I'll reintroduce yellow.

app/src/main/java/de/grobox/transportr/utils/FullScreenUtil.kt
newhinton: Ahh, sorry, typo. Fixed.

app/src/main/java/de/grobox/transportr/utils/FullScreenUtil.kt
newhinton: I will test if that creates the issues. 

It wasn't nessessary before because the content wasn't allowed to go behind the status bar. But since we now have the map behind the status bar, we need to move the actual content (eg. the back button) back down a bit, otherwise it will also be behind the bar. We could use a static margin, but since the status bar can have different heights depending on notches, we have to calculate it on demand.

There might be a better way, but if there is, i am not aware of it. It also should only affect the choosen view, not the rest.

Edit:

Disabling that code does not affect the behaviour of the bottomsheets.

app/src/main/res/drawable/ic_launcher_foreground_primary_color.xml
newhinton: Sadly no. It's used in the Theme.App.Starting-Style for the new Splash screen, where you cant set a tint, afaik.

https://developer.android.com/develop/ui/views/launch/splash-screen#splash-screen-resources

app/src/main/res/layout/fragment_directions_form.xml
newhinton: We dont? We use the default text color here, which is a dark/light grey depending on the current device theme.

app/src/main/res/layout/list_item_favorite.xml
newhinton: This is different than the above one. Here we actually re-set the text color via the FavoriteCardForeground-style. That sets  the text color. Removing the coloring makes it look like this:

image
We will have to set the text color anyway to fix the tinting. What color would you like?

newhinton:
> DatePicker dialog has different shape

This is actually not a regression, it is also present in the current design. It is just a bit more visible in this PR.

> Personally, I'm not fully convinced on giving up on the nice Transportr red altogether, given that it's part of the branding, but it seems that's what Material You is suggesting to do?

Yes it is. Though i recommend not to introduce a toggle ability, it would require quite a bit of code to make it work properly. It will lead to bugs. And i don't think i want to spend time on implementing that, especially since there is still a bit to do with the base-proposal.

newhinton:
Todo:
- [ ] Work through code annotations
- [x] Popup Menu - no icons
- [x] Network-Selection coloring
- [x] Mapview Backbutton
- [ ] Map-attribution behind bottombar
- [ ] transport mode selection dialog theme
- [ ] Select Location from main searchbar, dismiss bottom sheet. Re-select location. Left side of bottom sheet has padding
- [ ] Drop shadow of navigation also has padding after above flow
- [ ] Navigation Drawer is invisible below status bar when opened
- [ ] Three State Bottom Sheet has buggy behaviour
- [ ] Trip Details toolbar menu does not show up



i dont think its a Todo:

- [ ] Datepicker-style
- [ ] Select Location on map. Start navigation via FAB. New Activity has favorite locations instead of empty list (already present in 2.2.1)

.idea/codeStyles/Project.xml
```diff
@@ -36,9 +36,6 @@
     <JetCodeStyleSettings>
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </JetCodeStyleSettings>
-    <XML>
-      <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
-    </XML>
```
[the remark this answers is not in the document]
newhinton: @ialokim I have never (manually) touched those files. Should IDE-config files be in this repository anyhow?

newhinton:
@ialokim @Altonss 

I have improved on many of the things you noted. However, there are still some things broken, like the three-state-bottomsheet.

I also found that some issues are very likely due to older libraries beeing used. That opened an entire different can of worms. As an example: I have tried to manually calculate the statusbar height so that i can adjust the backButton in the tripview, with less than optimal results. In newer libraries there is an `enableEdgeToEdge()` function for Activity. I tried updating the dependency, but that also requires an update to kotlin, which deprecates `kotlin-android-extensions` which means that we have to switch to viewbindings. 

This is rapidly exploding the size of this PR, but i also don't think that we should continue without upgrading the libraries, which are at this point 2-4 years old and up to 5 versions behind. 

How should i continue?

PR #930, opened 2024-02-04
Add newhinton to CLA

Altonss:
Agreed here: https://github.com/grote/Transportr/pull/927#issuecomment-1919507873
cc @newhinton


michaelblyons:
After merging, can you hide or delete all the Bot messages in the other ticket?

PR #935, opened 2024-02-26
README: thanking JawgMaps for free vector map tile services

ialokim:
We could also think about crediting them a bit more prominent within the app. Currently, we only display the mandatory copyright notice on the map view itself.


Joxit:
Hi, thank you for adding Jawg Maps in your readme :)

Issue #970, opened 2024-12-02
Remove Transport NSW (Australia) network?


Altonss:
Transport NSW network seems to be broken (see https://github.com/grote/Transportr/pull/969#issuecomment-2512039743)

As seen in https://github.com/schildbach/public-transport-enabler/issues/252#issuecomment-2321788083, this appears to be due to the API, which seems to have been shut down for a while now.

So the simplest solution to this issue might be to remove this broken network.

Issue #939, opened 2024-03-25
IndexOutOfBoundsException in DepartureAdapter


grote:
``` ruby
5. Exception java.lang.IndexOutOfBoundsException: Asked to get item at -1 but size is 0
  at androidx.recyclerview.widget.SortedList.get (SortedList.java:706)
  at de.grobox.transportr.departures.DepartureAdapter.getItem (DepartureAdapter.java:96)
  at de.grobox.transportr.departures.DeparturesActivity.loadMoreDepartures (DeparturesActivity.java:238)
  at de.grobox.transportr.departures.DeparturesActivity.lambda$onCreate$0 (DeparturesActivity.java:127)
  at de.grobox.transportr.departures.DeparturesActivity.$r8$lambda$3f5aVX6MH5h4XzeWB06Qw6Y12cc (DeparturesActivity.java)
  at de.grobox.transportr.departures.DeparturesActivity$$ExternalSyntheticLambda0.onRefresh (R8$$SyntheticClass)
  at com.omadahealth.github.swipyrefreshlayout.library.SwipyRefreshLayout$1.onAnimationEnd (SwipyRefreshLayout.java:176)
  at com.omadahealth.github.swipyrefreshlayout.library.CircleImageView.onAnimationEnd (CircleImageView.java:108)
```

Issue #934, opened 2024-02-08
BEG (BayernProvider) has wrong/missing description


BloodPawHD:
**Describe the bug**
I opened an issue in the public-tranport-enabler repo (https://github.com/schildbach/public-transport-enabler/issues/560) because of a bug in the MVV network and  developer Schildbach reminds me to use the BayernProvider instead, because it works for Bavaria, that means  for Munich, too. BayernProvider is the public-transport-enabler implementation of DEFAS, which is the public transport system of Bavaria (https://www.bayerninfo.de/en/about-bayerninfo-1/data-offer/public-transport-data). The Problem is that this isn't really intuitive, because the description in "app/src/main/res/values/strings.xml" line 322 is just "Bavaria, Würzburg, Regensburg". Users like me maybe think it's just for the area Würzburg and Regensburg and use worse Providers like MVV for Munich. Maybe it's time to drop the support of MVV and other Providers that are already covered in BayernProvider, but I already mentioned this in the public-transport-enabler issue.

**Expected behavior**
Idk but something like "Most of Germany" as in the Deutsche Bahn provider, for example "Most of Bavaria" or "Includes AVV, VAB, VLC, DING, VGI, VVM, MVV, VGN, VLP, RVV, VGRI, bodo, RMV, OAM, VRN, SVV, VVT" ( a whole list of transport networks supported by DEFAS are listed  in their PDF https://www.bayern-fahrplan.de/de/de/media/files/liniennetzplan-2024.pdf).

M00NJ:
After the MVV provider being completely [broken](https://github.com/schildbach/public-transport-enabler/issues/567) the [PR](https://github.com/schildbach/public-transport-enabler/pull/414) for changing the endpoints is now finally being merged. This will also bring real time data to the MVV network. There is also an advantage to a provider that covers a smaller area, since you get fewer suggestions when searching a location. Just look up "Marienplatz" with the BEG provider. There are 45 suggestions when I only need 1 since I live in Munich. So I'm against dropping the MVV provider. But changing the description makes sense.

Issue #897, opened 2023-12-08
Bottom trip banner can disappear after swipe down of trip details view


Altonss:
**Describe the bug**
The bottom trip banner can disappear after swipe down of trip details view, leaving the screen with only a blank banner at the bottom, without a possibility to bring up the trip details again.

**To Reproduce**
1. Click on a specific trip to show the detailed view.
2. Swipe down the details to show the map.
3. Repeat these swipes several times completely (reducing trip info to bottom banner).
4. See error: sometimes the banner becomes blank, leaving no other possibility then to go back.

**Expected behavior**
Do not leave the bottom banner blank.

**Screenshots**
![image](https://github.com/grote/Transportr/assets/66519591/5b7cc849-cda9-4400-a08a-308ed3ba659c)

**Versions (please complete the following information):**
 - **Transportr Version:** 2.1.5 and 2.2.0 (preview)
 - **Android Version:** 13

newhinton:
The culprit is likely the handling of bottom-sheets. It also applies to the `activity_map` bottom sheet, aswell as the `fragment_trip_detail` bottom sheet.

> In the TripDetailFragment, the bottomsheet can be closed by swiping down the topBar. When swiping down any other element, it collapses properly.

I think they are not meant to be used in this way. They have a `behavior_hideable` attribute, but if they are set to false, the bottom sheets dont resize and always take their full space, which is also unwanted.

woheller69:
fixed in v2.2.5

PR #999, opened 2025-11-16
* Remove broken Sydney provider * Add clear all button for stored trips * Minor bugfixes

woheller69:
#970 
#939 
#934 
#897 
#797 

prepare for v2.2.5

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -8,8 +8,8 @@ apply from: 'witness.gradle'
 android {
 
 	defaultConfig {
-		versionCode 128
-		versionName "2.2.4"
+		versionCode 129
+		versionName "2.2.5"
 
 		applicationId "de.grobox.liberario"
 		minSdkVersion 24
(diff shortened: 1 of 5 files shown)


Echolon:
I just checked from a plain new install the recent two release changes:

- No Australia nor South America transportation networks offered anymore.
- Clear all button for stored trips works
- DB, VRS, VVS works
- Poland works
- On the bugs I'm not really sure how to test.
- Does location update button mean, there is a button on the left that I can press so it searches again? If so, yes that works.
- Alert seem to be possible. But the context menu icons are very left bound to the context menu corner. Is that normal?

Somehow now also DB prices work now for regional and long-distance. Great!

Though, I'm experiencing loading times - is it just me? That may be due to frequent service changes.

woheller69:
location button is in the FROM address field. If pressed the location is updated.

Context menu icons are as before, I only added 'Alert'

PR #1007, opened 2025-12-17
Replaced the link to Transifex

Kachelkaiser:


Kachelkaiser:
I only updated the link here in the contributing file. Maybe this has to be done on the website as well...

